PR #15, opened 2021-06-14
Add the singlejs script that bundles the Chime SDK for JavaScript int…

simmkyu:
…o a JavaScript file

**Issue #:**
N/A

**Description of changes:**
- Move the [singlejs demo](https://github.com/aws/amazon-chime-sdk-js/tree/master/demos/singlejs) from the Chime SDK for JavaScript repo to this sample repo.

**Testing**

1. How did you test these changes? I followed the steps to generate the JS file.
2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? No


utils/singlejs/.gitignore
```diff
@@ -0,0 +1,3 @@
+.DS_Store
```
ltrung, line 1: Do we need this? .DS_Store and build are already in the parent .gitignore.

utils/singlejs/README.md
```diff
@@ -0,0 +1,38 @@
+## Bundling Chime SDK into a single .js file
+
+This demo shows how to bundle the latest release of `amazon-chime-sdk-js` into a single JS file using Rollup. You'll need to have Node 10+ and npm 6.11+ on your local development machine.
```
ltrung, line 3: Node 10 is no longer supported. Can we update to Node 14 or 16?

utils/singlejs/.gitignore
simmkyu: It would be great to have the `.gitignore` per folder. For example, if we change the output folder from `build` to `dist`, we can modify only the `.gitignore` in the project.

utils/singlejs/README.md
simmkyu: Updated it to `14+`. Thanks for catching!

ltrung (approved):

PR #21, opened 2021-07-01
Add fs-extra dependency

jing-chen1:
**Issue #:** N/A

**Description of changes:**
Coming across `lack of module fs-extra` issue when running the deploy script of meeting demo. So add `fs-extra` into meeting demo dependencies

We do `const fs = require('fs-extra')` in the [deploy.js](https://github.com/aws-samples/amazon-chime-sdk/blob/main/apps/meeting/serverless/deploy.js#L2). But we don't have `fs-extra` in the package.json/pacage-lock.json of the meeting demo. Previously, the meeting demo is within React component library, and the `fs-extra` is in the package-lock.json of React library, so it works. But for now, we move the meeting demo out, so need to add `fs-extra` seperately.

**Testing**

1. How did you test these changes? run deploy script under serverless folder
2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? Yes, it can be tested using meeting demo


apps/meeting/package.json
```diff
@@ -11,6 +11,7 @@
     "amazon-chime-sdk-component-library-react": "^2.6.0",
     "amazon-chime-sdk-js": "^2.11.0",
     "aws-sdk": "^2.731.0",
+    "fs-extra": "^10.0.0",
```
ltrung, line 14: Where does this come from?

jing-chen1: This is in the [deploy.js](https://github.com/aws-samples/amazon-chime-sdk/blob/main/apps/meeting/serverless/deploy.js#L2) and we do `const fs = require('fs-extra')`. Also updated the PR description

simmkyu: Probably, we missed it because the servereless package uses the `fs-extra` module defined in the root level.
https://github.com/aws/amazon-chime-sdk-js/blob/917c330a3fc6e8d2349162b00d5015f36d4edf78/package.json#L68

ltrung (approved):

PR #36, opened 2021-08-11
Implement VideoGridProvider and GridControl component

xuesichao:
**Issue #:**
Implement the gallery view in meeting demo

**Description of changes:**
* Use `priorityBasedDownlinkPolicy` and enable `simulcast` in `meetingManager`
* [Unfinalized] Add new provider `VideoGridProvider` to support zoom in, zoom out and pagination related functions
  * Add a hook `useVideoGridState` to expose:
    * Attendee related state: `attendees`, `availableVideoSources`
    * Pagination related state: `currentPage`, `isZoomed`, `zoomedLevel` and `zoomedLevelIndex`
  * Add a hook `useVideoGridControls` to expose pagination related actions`zoomIn()`, `zoomOut()`, `prevPage()` and `nextPage()`
* [Temporary] Add new component `GridControl` to handle pagination related action
* [Temporary] Add new container `GridContols` to hold all `GridControl`
* Bump up the `amazon-chime-sdk-js` version from 2.13.0 to 2.15.0

**Testing**

Yes
Yes, meeting demo


apps/meeting/src/meetingConfig.ts
```diff
@@ -1,31 +1,20 @@
...
-
-const postLogConfig = {
-  name: 'SDK_LOGS',
-  batchSize: 85,
-  intervalMs: 2000,
-  url: `${BASE_URL}logs`,
-  logLevel: SDK_LOG_LEVELS.info
-};
+const simulcastEnabled = true;
+const logger = new ConsoleLogger('SDK', LogLevel.INFO);
```
ltrung, line 11: Cannot do console logger. It has to be the same Multilogger as before.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,319 @@
...
+export const initialState: State = {
+  attendees: {},
+  availableVideoSources: [],
+  currentPage: 1,
+  isZoomed: false,
+  zoomedLevel: 0,
+  zoomedLevelIndex: 0,
+};
+
+const zoomedLevels: number[] = [0, 1, 2, 4, 8, 16, 25];
```
ltrung, line 112: Do we need so many zoom levels?

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,319 @@
...
+  VideoPreferences,
+  VideoSource,
+} from 'amazon-chime-sdk-js';
+import { priorityBasedPolicy } from '../../meetingConfig';
+
+interface AttendeeStatus {
+  attendeeId: string;
+  name: string;
+  videoEnabled: boolean;
+  // sharingContent: boolean;
```
ltrung, line 13: Is this not needed?

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,319 @@
...
+  availableVideoSources: [],
+  currentPage: 1,
+  isZoomed: false,
+  zoomedLevel: 0,
+  zoomedLevelIndex: 0,
+};
+
+const zoomedLevels: number[] = [0, 1, 2, 4, 8, 16, 25];
+
+export function reducer(state: State, { type, payload }: Action): State {
```
ltrung, line 114: Can we break down this function? It is a bit too long.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,319 @@
...
+    zoomedLevel = zoomedLevels[zoomedLevelIndex];
+  };
+
+  const calculateVideoSourcesToBeRendered = (): string[] => {
+    const left = (currentPage - 1) * zoomedLevel;
+    const right = currentPage * zoomedLevel;
+    return availableVideoSources.slice(left, right);
+  };
+
+  const updateDownlinkPreferences = (): void => {
```
ltrung, line 142: Can we move these function out? This is too long to be nested function.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,319 @@
...
+      };
+    }
+    case VideoGridAction.ZoomIn: {
+      if (zoomedLevelIndex > 1) {
+        zoomedLevelIndex -= 1;
+        zoomedLevel = zoomedLevels[zoomedLevelIndex];
+        isZoomed = true;
+        updateDownlinkPreferences();
+      } else {
+        console.debug('Can not zoom in');
```
ltrung, line 247: You may want to use logger to log these as error.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,319 @@
...
+import { priorityBasedPolicy } from '../../meetingConfig';
+
+interface AttendeeStatus {
+  attendeeId: string;
+  name: string;
+  videoEnabled: boolean;
+  // sharingContent: boolean;
+}
+
+export type RosterAttendeeType = {
```
ltrung, line 16: Isn't this the same as in the library? Why we redefine it https://github.com/aws/amazon-chime-sdk-component-library-react/blob/819a2e6817659f4f02a3f7b16b58d128c854fc2f/src/types/index.ts#L50?

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,319 @@
...
+      return {
+        attendees,
+        availableVideoSources,
+        currentPage,
+        isZoomed,
+        zoomedLevel,
+        zoomedLevelIndex,
+      };
+    }
+    case VideoGridAction.UpdateRoster: {
```
ltrung, line 218: Could we rename this UpdateRoster to something else? It does not updateRoster just set video flag.

apps/meeting/src/providers/VideoGridProvider/state.tsx
xuesichao: Cause we didn't export this from our library

apps/meeting/src/providers/VideoGridProvider/state.tsx
xuesichao: Updated to `UpdateAttendeeStates`

apps/meeting/src/providers/VideoGridProvider/state.tsx
xuesichao: Here it works as a closure, it's easier to access the `state`. If we move it out, we need to pass the states to these functions, which could make code worse. But I will try to see which one better.

xuesichao: Moved these function out

apps/meeting/src/components/Icons/FeaturedLayout.tsx
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
```
LeviSklaroff, line 2: I think this demo is supposed to be changed from Apache-2.0 to MIT-0 as the repository is under MIT-0

xuesichao: Yeah, that's right, I need to fix that. I will push a new version to fix this when I rebase this commit.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+  videoSourceState: VideoSourceState;
+};
+
+export const initialState: State = {
+  gridState: {
+    mode: VideoGridMode.GalleryView,
+    isZoomed: false,
+    threshold: 8,
+  },
+  attendeeStates: {},
```
devalevenkatesh, line 45: Suggest changing `attendeeStates` to just `attendees`, or use `attendeeState` while looping over these down below in the reducer actions.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+        }
+      }
+
+      // Update the `videoEnabled` according to video sources
+      for (const attendeeId of videoSourceIds) {
+        if (!(attendeeId in attendeeStates)) {
+          if (isContentShare(attendeeId)) {
+            attendeeStates[attendeeId] = {
+              attendeeId,
+              name: 'content share',
```
devalevenkatesh, line 377: Any reason for the change in name here or keeping it '' (empty) in the else block?

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+        if (
+          isContentShare(attendeeId) &&
+          !videoSourceIds.includes(attendeeId)
+        ) {
+          delete attendeeStates[attendeeId];
+        }
+      }
+
+      // Update the `videoEnabled` according to video sources
+      for (const attendeeId of videoSourceIds) {
```
devalevenkatesh, line 372: The first loop where yyou map to get `videoSourceIds` can be avoided right? You can just use:
```js
for (const videoSource of videoSources) {
const attendeeId = videoSource.attendee.attendeeId;
..rest same.
}

File: apps/meeting/src/providers/VideoGridProvider/state.tsx [hunk 13]
```diff
@@ -0,0 +1,510 @@
...
+              name: '',
+              bandwidthConstrained: false,
+            } as AttendeeState;
+          }
+        }
+
+        attendeeStates[attendeeId].videoEnabled = true;
+      }
+
+      // Populate the `cameraSources` based on the order of attendeeStates
```
devalevenkatesh, line 392: nit: Add backtick for attendeeStates as well.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+      for (const attendee of Object.values(attendeeStates)) {
```
devalevenkatesh, line 393: Can we avoid this loop again and in the previous `for` itself keep updating the `cameraSources`:
```js
...
attendeeStates[attendeeId].videoEnabled = true;
cameraSources.push(attendeeId);
```

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+        gridState.isZoomed = false;
+        updateDownlinkPreferences(gridState, videoSourceState, attendeeStates);
+      }
+
+      return {
+        ...state,
+      };
+    }
+    default:
+      throw new Error('Incorrect type in VideoGridStateProvider');
```
devalevenkatesh, line 508: The filename says `VideoGridProvider`,  so need to update `VideoGridStateProvider` to the same.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+  | UpdateActiveSpeakers
+  | PauseVideoTile
+  | UnpauseVideoTile
+  | ZoomIn
+  | ZoomOut;
+
+const isContentShare = (sourceId: string): boolean =>
+  new DefaultModality(sourceId).hasModality(DefaultModality.MODALITY_CONTENT);
+
+const calculateVideoSourcesToBeRendered = (
```
devalevenkatesh, line 165: Should we make a new `Utils.ts` files in `VideoGridProvider/Utils.ts` and move `updateDownlinkPreferences`, `calculateVideoSourcesToBeRendered` and `isContentShare` there?

apps/meeting/src/providers/VideoGridProvider/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+      return;
+    }
+
+    dispatch({
+      type: VideoGridAction.UpdateAttendeeStates,
+      payload: { roster },
+    });
+  }, [audioVideo, roster]);
+
+  useEffect(() => {
```
devalevenkatesh, line 53: Suggest moving this `useEffect` to the top. React executes hooks in the order they appear in an component.

apps/meeting/src/providers/VideoGridProvider/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+    if (!audioVideo) {
+      return;
+    }
+
+    dispatch({
+      type: VideoGridAction.UpdateAttendeeStates,
+      payload: { roster },
+    });
+  }, [audioVideo, roster]);
```
devalevenkatesh, line 51: Could you please explain why does this need `audioVideo` dependency?

apps/meeting/src/providers/VideoGridProvider/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  useEffect(() => {
+    if (!audioVideo) {
+      return;
+    }
+
+    return (): void => dispatch({ type: VideoGridAction.ResetVideoGridState });
+  }, [audioVideo]);
+
+  useEffect(() => {
```
devalevenkatesh, line 61: Great job separating concerns in their own hooks 👍

apps/meeting/src/providers/VideoGridProvider/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+      type: VideoGridAction.UpdateLocalSourceState,
+      payload: {
+        isVideoEnabled,
+        localAttendeeId,
+        isLocalUserSharing,
+        sharingAttendeeId,
+      },
+    });
+  }, [
+    audioVideo,
```
devalevenkatesh, line 143: Here as well, any reason for the `audioVideo` dependency?

apps/meeting/src/components/Icons/Spinner.tsx
```diff
@@ -5,8 +5,6 @@ import React from 'react';
 import { faSpinner } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
-const Spinner: React.FC = () => {
-  return <FontAwesomeIcon icon={faSpinner} spin />;
-};
+const Spinner: React.FC = () => <FontAwesomeIcon icon={faSpinner} spin />;
```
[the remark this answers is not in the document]
xuesichao: Yes, I will fix this conflict when I do the rebase.

apps/meeting/src/containers/Navigation/index.tsx
```diff
@@ -9,37 +9,90 @@ import {
...
-      >
-        <LocalMediaStreamMetrics />
-      </NavbarItem>
+      <Flex css="margin-top: 0rem;">
+        <NavbarItem
+          icon={<Attendees />}
+          onClick={toggleRoster}
+          label="Attendees"
+        />
+        <NavbarItem
```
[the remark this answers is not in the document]
xuesichao: Nice catch. I'm actually following the Jaime's UX design. This button should be called "go to xxx mode". That's why it displayed GalleryLayout icon when it's actually Featured mode and vice verse.

I think another reason why it is confused is because the difference between two mode is not very significant, so user have to check something to tell what mode they are actually using. If you check [zoom's guide](https://support.zoom.us/hc/en-us/articles/201362323-Changing-the-video-layout-Active-Speaker-View-and-Gallery-View-#h_ac952845-ed43-4884-8263-4356f98e26f3), they use the same logic.

apps/meeting/src/providers/VideoGridProvider/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+  useLocalVideo,
+  useMeetingManager,
+  useRosterState,
+} from 'amazon-chime-sdk-component-library-react';
+import {
+  AudioVideoObserver,
+  VideoDownlinkObserver,
+  VideoSource,
+} from 'amazon-chime-sdk-js';
+import React, { createContext, useContext, useEffect, useReducer } from 'react';
```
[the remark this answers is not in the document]
xuesichao: Very good suggestions. I'm curious about the conversion of import statement. I will follow your advice.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+};
+
+type UpdateGridState = {
+  type: VideoGridAction.UpdateGridState;
+  payload: {
+    videoGridMode: VideoGridMode;
+  };
+};
+
+type UpdateAttendeeStates = {
```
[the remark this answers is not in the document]
xuesichao: At the beginning I called it `UpdateRoster`, but Trung left a comment and I changed its name to `UpdateAttendeeStates`.
The actual roster doesn't depend on this variable. This variable is used to store the info for each attendee or each video sources. It's not the actually roster itself, but they are quite similar. So I would prefer not to change this name.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+};
+
+type UpdateActiveSpeakers = {
+  type: VideoGridAction.UpdateActiveSpeakers;
+  payload: {
+    activeSpeakers: string[];
+  };
+};
+
+type UpdateGridState = {
```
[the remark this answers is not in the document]
xuesichao: There were more variable in this state before, like the threshold, the size of grid. That's why it's called  `UpdateGridState`, but I removed most of them and only a `gridMode` left. Should I keep the naming for the possibility in the future we add more grid state related variables?

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+    targetDisplaySize = TargetDisplaySize.Low;
+  }
+
+  for (const videoSource of videoSourcesToBeRendered) {
+    const { attendeeId, type } = videoSource;
+
+    // Prioritize Content Share
+    if (type === 'contentShare') {
+      videoPreferences.add(
+        new VideoPreference(attendeeId, 1, TargetDisplaySize.High)
```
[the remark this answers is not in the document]
xuesichao: Oh this 1 and 2 means the priority, the lower the number the higher the priority.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+          attendeeId,
+          1,
+          mode === VideoGridMode.FeaturedView
+            ? TargetDisplaySize.High
+            : targetDisplaySize
+        )
+      );
+    }
+
+    // Set the common tiles to low priority
```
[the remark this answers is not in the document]
xuesichao: `VideoPreference(attendeeId, priority, targetDisplaySize)`

The lower the value of priority (number) the higher the priority. The low priority video sources are first to be paused in a restriction network.
The `targetDisplaySize` is used to control the resolution of the video source.

Here we always set the common type `priority` to 2 (low). For content share and active speaker type we set it to 1 (high). So the comment should be fine.

Maybe I can have a `enum` or `type` for priority to make it more clear.

apps/meeting/src/providers/VideoGridProvider/state.tsx
xuesichao: Good catch!

apps/meeting/src/providers/VideoGridProvider/state.tsx
xuesichao: There's a racing condition between `remoteVideoSourcesDidChange` and `AttendeeIdPresence`. Here I use the `useRosterState` hook, where we skip the content share. So here, whenever we find the content share source via `remoteVideoSourcesDidChange`, we assign it name 'content share'.(Cause `useRosterState` skips the content share, and there is no way to update it's name, check [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/d1fc0f5312e8cc4f47ebceab030cbbd8bc7bdb90/src/providers/RosterProvider/index.tsx#L44-L47)). But for non-content-share video source their name can be updated via `AttendeeIdPresence` (`useRosterState`).

So it's better to use `audioVideo.realtimeSubscribeToAttendeeIdPresence(cb)` here instead of `useRosterState`, but I didn't realize this when I implement the project. I should improve the code in the future to make things simple.

apps/meeting/src/providers/VideoGridProvider/state.tsx
xuesichao: We need to check whether content share id is in the `videoSourceIds`. So have to keep it, but to optimize the performance, I update the `videoSourceIds` from array to set `videoSourceIdSet`.

apps/meeting/src/providers/VideoGridProvider/state.tsx
xuesichao: I'm not 100% sure, but I feel like the order matters here. The previous based on the order of id in `videoSources`.
This loop based on the order of `attendeeStates` (roster list).
I'm not sure if there's any relation between the order of remote tiles and the order of video preferences.
So have to keep it. If later we found the order doesn't matter here, we could optimize it, and other places.

apps/meeting/src/providers/VideoGridProvider/state.tsx
```diff
@@ -0,0 +1,510 @@
...
+
+
+  // Last, add common video sources
+  let gridSize = 0;
+
+  if (mode === VideoGridMode.GalleryView) {
+    if (isZoomed) {
+      gridSize = threshold;
+    } else {
+      gridSize = Number.MAX_SAFE_INTEGER;
```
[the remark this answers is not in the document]
xuesichao: Yes. 
If we keep it as a constant, the max number of video tiles could increase in the future, then we have to maintain it. I don't see value here so to make things simple, I will keep using the `Number.MAX_SAFE_INTEGER`

PR #49, opened 2021-09-01
Fixed Android-WebView-Sample dependency versioning issue

LeviSklaroff:
**Issue #:**
The Android WebView demo fails to build because of a versioning issue from a dependency as it was using latest version instead of current.
**Description of changes:**
Set a specific version for the dependency. Also swapped folder casing to lowercase to match naming convention.
**Testing**

- Installed Java 11
- Built the gradle project using ./gradlew build
No


ltrung:
Can we change the name of the demo to lowercase as discussed in our channel?

LeviSklaroff:
> Can we change the name of the demo to lowercase as discussed in our channel?

Changed

ltrung (approved):

PR #52, opened 2021-09-02
Add preinstall script to use latest component library from GitHub

devalevenkatesh:
**Issue #:**
- The meeting demo does not currently get the updated version of react component library and hence may run into build issues.

**Description of changes:**
- This adds a postinstall npm script to clone the react component library so that the meeting demo uses the latest code.

**Testing**

- Ran the script locally to check whether the meeting demo can be run with the new changes in Github.
- To do this, installed an older version of react library (2.8.0) and then ran `npm install` which updated the react library to the latest tarball generated with the current code in the react library master branch.
- Did sanity checks and all worked fine.
- React meeting demo.


apps/meeting/package.json
```diff
@@ -15,11 +15,13 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-router-dom": "^5.1.2",
+    "rimraf": "^3.0.2",
     "styled-components": "^5.0.0",
     "styled-system": "^5.1.5",
     "throttle-debounce": "^2.3.0"
   },
   "scripts": {
+    "postinstall": "node scripts/update-react-library.js",
```
ltrung, line 24: Is there a reason to use postinstall vs preinstall?

apps/meeting/scripts/update-react-library.js
```diff
@@ -0,0 +1,53 @@
...
+ * GitHub repository.
+ */
+
+const fs = require('fs');
+const path = require('path');
+const { logger, spawnOrFail } = require('./utilities');
+const rimraf = require('rimraf');
+const packageJSON = require('../package.json');
+
+const dir = 'updated-react-library';
```
ltrung, line 14: Can we call it what it is `amazon-chime-sdk-component-library-react-github`?

apps/meeting/scripts/update-react-library.js
```diff
@@ -0,0 +1,53 @@
...
+
+if (!fs.existsSync(dir)) {
+  logger.log(`Creating a temp directory ${dir}`);
+  fs.mkdirSync(dir);
+}
+
+process.chdir(path.join(__dirname, `../${dir}`));
+logger.log(`Cloning ${package} to update the existing installation.`);
+spawnOrFail('git', [`clone ${packageGitUrl}`]);
+logger.log('Cloning complete.');
```
ltrung, line 27: nit: Dont think we need this line.

apps/meeting/scripts/update-react-library.js
ltrung: or github packages if we want to maybe do js sdk in the future :D

apps/meeting/package.json
devalevenkatesh: Any works, just wanted to do it once the rest are installed. Do not think it really matters here, should I use pre instead?

apps/meeting/scripts/update-react-library.js
devalevenkatesh: Yeah makes sense, will update it thanks!

apps/meeting/scripts/update-react-library.js
devalevenkatesh: Ok, removed it 👍

apps/meeting/package.json
michhyun1: If I ever want to use the npm version, how would I do that?

If I do npm install amazon-chime-sdk-component-library-react@2.9.0

wouldnt the postinstall script run and just overwrite the amazon-chime-sdk-component-library-react dependency to point to github master?

devalevenkatesh: Yeah good question and I have tested it already. So, the `post` `pre` install scripts will only run on the `npm install` script and not `npm install <package>`.

PR #60, opened 2021-09-15
Split webpack to dev and prod to reduce the bundle size for serverless demo

ltrung:
**Description of changes:**
Running in dev with inline-source-map makes the bundle size bigger than the Lambda limit so serverless demo would not be able to fetch the page.
Create a new dev webpack config to run with npm run start for debugging. npm run build will use the default - production mode to reduce the bundle size.

**Testing**

- Test both `npm run start` and `npm run build` and verify the bundle size.
- Do serverless deployment and verify that it loads
2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? Yes, meeting demo.

### DIFF
--- a/apps/meeting/package.json
+++ b/apps/meeting/package.json
@@ -8,8 +8,8 @@
     "@types/webpack-env": "^1.15.1",
     "@typescript-eslint/eslint-plugin": "^2.19.2",
     "@typescript-eslint/parser": "^2.19.2",
-    "amazon-chime-sdk-component-library-react": "^2.9.0",
-    "amazon-chime-sdk-js": "^2.16.1",
+    "amazon-chime-sdk-component-library-react": "^2.9.1",
+    "amazon-chime-sdk-js": "^2.17.0",
     "aws-sdk": "^2.731.0",
     "fs-extra": "^10.0.0",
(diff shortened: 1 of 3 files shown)


apps/meeting/package.json
```diff
@@ -8,8 +8,8 @@
     "@types/webpack-env": "^1.15.1",
     "@typescript-eslint/eslint-plugin": "^2.19.2",
     "@typescript-eslint/parser": "^2.19.2",
-    "amazon-chime-sdk-component-library-react": "^2.9.0",
-    "amazon-chime-sdk-js": "^2.16.1",
+    "amazon-chime-sdk-component-library-react": "file:../../../../amazon-chime-sdk-component-library-react/amazon-chime-sdk-component-library-react-2.9.1.tgz",
```
xuesichao, line 11: Did you make this change particularly?
Feels like we should change it to 2.9.1.

simmkyu: A nice catch

ltrung: Oops, remnant from my previous testing with react library :D

xuesichao (approved):

simmkyu (approved):

PR #65, opened 2021-09-22
Enable source map loader for meeting demo

ltrung:
**Issue #:**
Currently when running in dev mode, we do not load the source map for component library which makes it hard for debugging.
**Description of changes:**
Enable source map loader for react library in webpack.
The react library already contains source map file.

**Testing**

1. How did you test these changes? `npm run start` and verified that we could put break point in component library.
2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? Yes, meeting demo.

### DIFF
--- a/apps/meeting/webpack.config.dev.js
+++ b/apps/meeting/webpack.config.dev.js
@@ -1,3 +1,20 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT-0
+
 const config = require('./webpack.config.js');
 
-module.exports = Object.assign(config, { mode: 'development', devtool: 'inline-source-map' });
+module.exports = Object.assign(config, { 
+  mode: 'development', 
+  devtool: 'inline-source-map', 
+  module: {
+    rules:  [
+      ... config.module.rules,
+      {
+        test: /\.js$/,
+        enforce: 'pre',
+        use: ['source-map-loader'],
+        include: /amazon-chime-sdk-component-library-react/
+      },
+    ]
+  }
+});
--- a/apps/meeting/package-lock.json
+++ b/apps/meeting/package-lock.json
@@ -14,13 +14,14 @@
         "@types/webpack-env": "^1.15.1",
         "@typescript-eslint/eslint-plugin": "^2.19.2",
         "@typescript-eslint/parser": "^2.19.2",
-        "amazon-chime-sdk-component-library-react": "^2.9.0",
-        "amazon-chime-sdk-js": "^2.16.1",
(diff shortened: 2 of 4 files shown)


apps/meeting/webpack.config.dev.js
```diff
@@ -1,3 +1,91 @@
-const config = require('./webpack.config.js');
+// Copyright 2020-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
simmkyu, line 1: Remove `2020-`?

apps/meeting/webpack.config.dev.js
```diff
@@ -1,3 +1,91 @@
...
+        test: /\.css$/,
+        use: ['style-loader', 'css-loader']
+      },
+      {
+        test: /\.(svg)$/,
+        type: 'asset/inline'
+      },
+      {
+        test: /\.js$/,
+        enforce: "pre",
```
simmkyu, line 31: Use single quotes

apps/meeting/webpack.config.dev.js
```diff
@@ -1,3 +1,91 @@
...
+      {
+        test: /\.css$/,
+        use: ['style-loader', 'css-loader']
+      },
+      {
+        test: /\.(svg)$/,
+        type: 'asset/inline'
+      },
+      {
+        test: /\.js$/,
```
simmkyu, line 30: No JSX files?

apps/meeting/webpack.config.dev.js
```diff
@@ -1,3 +1,91 @@
...
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+const app = 'meeting';
+
+module.exports = {
+  mode: 'development',
+  entry: ['./src/index.tsx'],
+  devtool: 'inline-source-map',
+  module: {
+    rules: [
```
simmkyu, line 15: It would be great if we could reuse common rules, resolve, and other fields between dev and prod.

apps/meeting/webpack.config.dev.js
ltrung: Update.

apps/meeting/webpack.config.dev.js
ltrung: Fixed.

apps/meeting/webpack.config.dev.js
ltrung: Component library compiles everything to js files and not jsx so I only match with js files.

simmkyu (approved):

devalevenkatesh (approved):

PR #70, opened 2021-10-11
Add reselect function

jing-chen1:
**Issue #:** Amazon Voice Focus does not get applied on new audio-devices when connected mid-meeting.

**Description of changes:** This is for the React library changes: [#639](https://github.com/aws/amazon-chime-sdk-component-library-react/pull/639). Implement the reselect function in meeting demo to add Amazon Voice Focus to a normal audio input device when devices change.

**Testing**

 * Open two Chrome tabs (no Safari, no FF)
 * Disable speaker in one tab
 * Disable microphone and set built in device as speaker in another tab
 * Turn on Amazon Voice Focus in the first tab
 * Apply a new device (AirPods)
 * The current device changes to AirPods
 * Then make some background noice, you should not hear any noice from the second tab since Amazon Voice Focus is enabled in the first tab.
 
Yes, in the meeting demo

### DIFF
--- a/apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
+++ b/apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
@@ -4,8 +4,14 @@
 import React from 'react';
 import { Route, Switch } from 'react-router-dom';
 import {
+  AudioTransformDevice,
+  Device,
+  VoiceFocusTransformDevice,
+} from 'amazon-chime-sdk-js';
+import {
   BackgroundBlurProvider,
   MeetingProvider,
+  useVoiceFocus,
   VoiceFocusProvider,
 } from 'amazon-chime-sdk-component-library-react';
 
@@ -17,6 +23,28 @@ import MeetingEventObserver from '../MeetingEventObserver';
 import meetingConfig from '../../meetingConfig';
 import { useAppState } from '../../providers/AppStateProvider';
 
+const MeetingProviderWithDeviceReplacement: React.FC = ({ children }) => {
+  const { addVoiceFocus } = useVoiceFocus();
+
+  const onDeviceReplacement = (
+    nextDevice: string,
+    currentDevice: Device | AudioTransformDevice,
+  ): Promise<Device | VoiceFocusTransformDevice> => {
+    if (currentDevice instanceof VoiceFocusTransformDevice) {
+      return addVoiceFocus(nextDevice);
+    }
+    return Promise.resolve(nextDevice);
+  };
+
+  const meetingConfigValue = {
+    ...meetingConfig,
+    enableWebAudio: true,
+    onDeviceReplacement,
+  };
+
+  return <MeetingProvider {...meetingConfigValue}>{children}</MeetingProvider>
+};
+
 const MeetingProviderWrapper: React.FC = () => {
   const { isWebAudioEnabled, isBackgroundBlurEnabled } = useAppState();
 
@@ -27,7 +55,7 @@ const MeetingProviderWrapper: React.FC = () => {
 
   const getMeetingProviderWrapper = () => {
     return (
-      <MeetingProvider {...meetingConfigValue}>
+      <>
         <NavigationProvider>
           <Switch>
             <Route exact path={routes.HOME} component={Home} />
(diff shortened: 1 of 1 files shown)


apps/meeting/src/app.tsx
```diff
@@ -57,6 +42,44 @@ const App: FC = () => (
...
+  const reselection = (device: Device): Promise<Device | VoiceFocusTransformDevice> => {
+    if (isVoiceFocusChecked) {
+      return addVoiceFocus(device);
+    }
+    return Promise.resolve(device);
+  };
+
+  const toggleVoiceFocus = () => {
+    setIsVoiceFocusChecked(current => !current);
+  }
```
richnew10, line 59: Remember our previous discussions about how this kind of state management isn't accurate, because applying Voice Focus can fail?

apps/meeting/src/app.tsx
```diff
@@ -57,6 +42,44 @@ const App: FC = () => (
...
+  const { addVoiceFocus } = useVoiceFocus();
+  const { meetingMode } = useAppState();
+  const [isVoiceFocusChecked, setIsVoiceFocusChecked] = useState(false);
+
+  const reselection = (device: Device): Promise<Device | VoiceFocusTransformDevice> => {
+    if (isVoiceFocusChecked) {
+      return addVoiceFocus(device);
+    }
+    return Promise.resolve(device);
+  };
```
richnew10, line 55: Why is this not the same kind of code that's in the input control, using the meeting manager to find out if VF is enabled, and preserving that across reselection?

There is definitely going to be duplicate code, so I wonder if we can break it out as a context?

richnew10 (changes requested):

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
```diff
@@ -17,6 +23,29 @@ import MeetingEventObserver from '../MeetingEventObserver';
...
+    if (currentDevice instanceof VoiceFocusTransformDevice) {
+      return addVoiceFocus(nextDevice);
+    }
+    return Promise.resolve(nextDevice);
+  };
+
+  const meetingConfigValue = {
+    ...meetingConfig,
+    enableWebAudio: isWebAudioEnabled,
+    onDeviceReplacement: isWebAudioEnabled ? replaceDevice : undefined
```
richnew10, line 43: If we never support turning off web audio — and I suggest we should not, because it won't work — I think you can simplify this. You check for web audio when deciding which provider to build, so you don't have to inside the provider.

richnew10 (changes requested):

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
jing-chen1: Good catch! Will set `true` to `enableWebAudio`, and `replaceDevice` to `onDeviceReplacement`

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
```diff
@@ -3,9 +3,15 @@
 
 import React from 'react';
 import { Route, Switch } from 'react-router-dom';
+import {
+  AudioTransformDevice,
+  Device,
+  VoiceFocusTransformDevice
```
richnew10, line 9: Nit: trailing commas, please.

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
```diff
@@ -17,6 +23,28 @@ import MeetingEventObserver from '../MeetingEventObserver';
...
+    if (currentDevice instanceof VoiceFocusTransformDevice) {
+      return addVoiceFocus(nextDevice);
+    }
+    return Promise.resolve(nextDevice);
+  };
+
+  const meetingConfigValue = {
+    ...meetingConfig,
+    enableWebAudio: true,
+    onDeviceReplacement: replaceDevice
```
richnew10, line 42: Trailing comma.

If you call `replaceDevice` `onDeviceReplacement` you can simplify this line.

richnew10 (approved):
Approved with nits.

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
jing-chen1: Sure, updated it

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
jing-chen1: Rename the function, it is simplified now :)

richnew10 (approved):

devalevenkatesh:
Closing as this is reviewed, merged in meeting-dev and main now.

PR #74, opened 2021-10-19
Add BackgroundBlurDeviceProvider and VideoInputBackgroundBlurControl to meeting demo

michhyun1:
**Issue #:**
https://github.com/aws/amazon-chime-sdk-js/issues/1690

**Description of changes:**
Add the `BackgroundBlurDeviceProvider` to the main component tree and `VideoInputBackgroundBlurControl` to the `MeetingControls`.

This depends on the 2.11.0 version of the Amazon Chime React SDK Component Library and also the 2.20.0 version of the Amazon Chime SDK for Javascript. Both have not been released yet, so this PR will sit in PR until both are released. 

**Testing**

```
1. npm run start
2. join a meeting
3. toggle video on and off
4. enable bg blur
5. toggle video on and off
6. switch between videos
```

### DIFF
--- a/apps/meeting/src/containers/MeetingControls/index.tsx
+++ b/apps/meeting/src/containers/MeetingControls/index.tsx
@@ -6,12 +6,13 @@ import {
   ControlBar,
   AudioInputVFControl,
   AudioInputControl,
-  VideoInputControl,
+  VideoInputBackgroundBlurControl,
   ContentShareControl,
   AudioOutputControl,
   ControlBarButton,
   useUserActivityState,
   Dots,
+  VideoInputControl,
 } from 'amazon-chime-sdk-component-library-react';
 
 import EndMeetingControl from '../EndMeetingControl';
@@ -22,7 +23,7 @@ import { useAppState } from '../../providers/AppStateProvider';
 const MeetingControls: React.FC = () => {
   const { toggleNavbar, closeRoster, showRoster } = useNavigation();
   const { isUserActive } = useUserActivityState();
-  const { isWebAudioEnabled } = useAppState();
+  const { isWebAudioEnabled, isBackgroundBlurEnabled } = useAppState();
 
   const handleToggle = (): void => {
     if (showRoster) {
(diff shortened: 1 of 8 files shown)


package-lock.json
```diff
@@ -1,6 +0,0 @@
-{
-  "name": "amazon-chime-sdk",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {}
-}
```
michhyun1, line 6: why is there a package-lock on the root level of this repo? We can delete it like this right?

apps/meeting/src/containers/MeetingControls/index.tsx
```diff
@@ -6,7 +6,7 @@ import {
   ControlBar,
   AudioInputVFControl,
   AudioInputControl,
-  VideoInputControl,
+  VideoInputBackgroundBlurControl,
```
devalevenkatesh, line 9: We do not want to remove the existing `VideoInputControl`, but provide a check to use background blur feature like we do for VF now.

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
```diff
@@ -26,24 +27,26 @@ const MeetingProviderWrapper: React.FC = () => {
 
   const getMeetingProviderWrapper = () => {
     return (
-      <MeetingProvider {...meetingConfigValue}>
```
devalevenkatesh, line 29: We do not want to change this right, but, add another method like `getMeetingProviderWrapperWithVF` say `getMeetingProviderWrapperWithBgBlur`. We want to test the demo such that we should be able to test demo with only bg blur, only VF, with VF and bgblur and without anything with simple controls without these features.

devalevenkatesh (changes requested):

apps/meeting/src/containers/MeetingControls/index.tsx
michhyun1: done

PR #94, opened 2021-11-18
Conditionally show the background blur checkbox depending on if background blur is enabled

michhyun1:
**Issue #:**

**Description of changes:**
Fix a bug where the background blur checkbox was being shown even when background blur was disabled.

**Testing**
ran the test locally and verified the checkbox shows when bg blur is enabled, and doesnt show when disabled. 

From the meeting form, disable bg blur, join the meeting, validate that the checkbox is not mounted. 
From the meeting form, enabled bg blur, join the meeting, validate that the checkbox is mounted.

React meeting demo

### DIFF
--- a/apps/meeting/src/components/DeviceSelection/CameraDevices/index.tsx
+++ b/apps/meeting/src/components/DeviceSelection/CameraDevices/index.tsx
@@ -12,8 +12,11 @@ import {
 } from 'amazon-chime-sdk-component-library-react';
 
 import { title, StyledInputGroup } from '../Styled';
+import { useAppState } from '../../../providers/AppStateProvider';
+import { BlurValues } from '../../../types';
 
 const CameraDevices = () => {
+  const { blurOption } = useAppState();
   return (
     <div>
       <Heading tag="h2" level={6} css={title}>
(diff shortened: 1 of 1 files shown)


ltrung (approved):

apps/meeting/src/components/DeviceSelection/CameraDevices/index.tsx
```diff
@@ -25,9 +28,11 @@ const CameraDevices = () => {
       <StyledInputGroup>
         <QualitySelection />
       </StyledInputGroup>
+      { blurOption !== BlurValues.blurDisabled ?
       <StyledInputGroup>
         <BackgroundBlurCheckbox />
-      </StyledInputGroup>
+      </StyledInputGroup> : ''
```
devalevenkatesh, line 34: Minor can fix later but Instead of "" can we just render like below (dont know what "" shows like 😅 ) : 
```
{ blurOption !== BlurValues.blurDisabled &&
      <StyledInputGroup>
        <BackgroundBlurCheckbox />
      </StyledInputGroup>}
```

Issue #80, opened 2021-10-27
Bug on livereload


Brenerss:
In the current amazon chime project, where the meeting folder is now separate, live reloading doesn't work, we need to restart the server all the time

xuesichao:
Hi @Brenerss , thanks for reporting this issue, I confirm it exists. After making some changes, the App reloads infinitely. Will look into this issue and post the update here.

Brenerss:
Ok, thanks

xuesichao:
Merged the fix to meeting-dev branch. The fix will be merged into main branch after next release.

PR #97, opened 2021-11-18
Fix webpack infinite reloading issue

xuesichao:
**Issue #80 :**

**Description of changes:**
* Remove `InlineChunkHtmlPlugin` which cause the infinite reloading issue and its dependency `react-dev-utils`.
* Remove the logic of sending HTML page from `server.js`.
  * `webpack-dev-server` now handles the rendering of HTML pages.
  * `server.js` handles Chime API calls only. 
* Add some missing trailing commas.
* Add boundary of webpack-dev-server proxy.

**Testing**

Run the meeting demo and make sure the reloading function is good.

meeting demo.

### DIFF
--- a/apps/meeting/webpack.config.js
+++ b/apps/meeting/webpack.config.js
@@ -38,32 +38,34 @@ module.exports = {
     fallback: {
       fs: false,
       tls: false,
-    }
+    },
   },
   output: {
     path: path.join(__dirname, 'dist'),
     filename: `${app}-bundle.js`,
     publicPath: '/',
     libraryTarget: 'var',
-    library: `app_${app}`
+    library: `app_${app}`,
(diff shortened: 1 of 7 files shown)


apps/meeting/webpack.config.js
```diff
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: MIT-0
 
 const path = require('path');
-const InlineChunkHtmlPlugin = require('react-dev-utils/InlineChunkHtmlPlugin');
```
devalevenkatesh, line 5: I think we cannot remove this plugin cause this is the one helping us putting all in one HTML file which we then deploy. Can you once check if this still helps us deploy using existing setup?

PR #103, opened 2021-11-20
 Fix the wrong display style of Zoom In/ Zoom Out label

xuesichao:
**Issue #:**

**Description of changes:**
Bug: Changing the CSS `display` to `none` only hides the icon itself, the label still displays.
Fix: Instead of changing the CSS `display` to `none`, change the Zoom In/Zoom Out buttons to be rendered conditionally.

**Testing**

Run the meeting demo and confirm the fix works.

Yes, meeting demo.


ltrung:
Can we add description of the actual bug?

xuesichao:
Close the PR cause it is already merged into meeting-dev branch

PR #108, opened 2021-12-02
Add Amazon Chime Echo Reduction for React SDK

wchmzn:
**Issue #:**

**Description of changes:**
Add Amazon Chime Echo Reduction for React SDK
**Testing**

1. How did you test these changes? Same as React lib changes: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/689
apps/meeting


devalevenkatesh (commented):
You need to update the PR with just 1 echo reduction commit, currently I see few more un-related commits.

apps/meeting/server.js
```diff
@@ -15,18 +15,34 @@ let protocol = 'http';
 let options = {};
 
 const chime = new AWS.Chime({ region: 'us-east-1' });
-const alternateEndpoint = process.env.ENDPOINT;
```
devalevenkatesh, line 18: We would still need this if we were to test the Chime's other (gamma) endpoint right?

apps/meeting/server.js
```diff
@@ -15,18 +15,34 @@ let protocol = 'http';
...
+  console.log('Using global endpoint ');
   chime.endpoint = new AWS.Endpoint(
     'https://service.chime.aws.amazon.com/console'
   );
 }
 
+// return regional API just for Echo Reduction for now.
+function getClientForMeeting(meeting) {
+  if (
+    meeting &&
```
devalevenkatesh, line 35: You can avoid this by using optional chaining feature in JS:
```
meeting &&
    meeting.Meeting &&
    meeting.Meeting.MeetingFeatures &&
    meeting.Meeting.MeetingFeatures.Audio &&
    meeting.Meeting.MeetingFeatures.Audio.EchoReduction === 'AVAILABLE'
```
to
```
meeting?.Meeting?.MeetingFeatures?.Audio?.EchoReduction === 'AVAILABLE'
```

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

apps/meeting/server.js
```diff
@@ -60,24 +76,34 @@ const server = require(protocol).createServer(
         const name = query.name;
         const region = query.region || 'us-east-1';
 
+        let client = getClientForMeeting(meetingCache[title]);
```
devalevenkatesh, line 79: nit: client can be a `const`.

devalevenkatesh, line 79: This is not useful if query param has `query.ns_es === 'true'`, so can this be moved to else block?

apps/meeting/server.js
```diff
@@ -60,24 +76,34 @@ const server = require(protocol).createServer(
...
           JoinInfo: {
             Title: title,
             Meeting: meetingCache[title].Meeting,
             Attendee: (
-              await chime
+              await client
                 .createAttendee({
                   MeetingId: meetingCache[title].Meeting.MeetingId,
-                  ExternalUserId: uuid()
+                  ExternalUserId: `${uuidv4().substring(0, 8)}#${query.name}`.substring(0, 64),
```
devalevenkatesh, line 106: [Understanding]: What is the purpose of this change?

devalevenkatesh (changes requested):
Have left some comments, request you to please cleanup last comments to review further, thanks.

apps/meeting/server.js
wchmzn: add back

apps/meeting/server.js
wchmzn: fixed

apps/meeting/server.js
wchmzn: i keep it the same as https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/server.js#L74-L91. client is used to createMeeting and createAttendee. which else block are you referring to?

apps/meeting/server.js
wchmzn: change back

apps/meeting/server.js
wchmzn: i will have to put it back because `?.` generates the following error
```
SyntaxError: Unexpected token '.'
     at wrapSafe (internal/modules/cjs/loader.js:1054:16)
     at Module._compile (internal/modules/cjs/loader.js:1102:27)
     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
     at Module.load (internal/modules/cjs/loader.js:986:32)
     at Function.Module._load (internal/modules/cjs/loader.js:879:14)
     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
     at internal/main/run_main_module.js:17:47
```

apps/meeting/server.js
wchmzn: i will have to change it back as the error `server caught error: ReferenceError: uuid is not defined`. I believe this is because we use regional endpoint here.

PR #113, opened 2021-12-18
Allow to enable simulcast and priority video policies

ltrung:
**Description of changes:**
- Allow to select simulcast uplink and priority-based downlink policies
- Disable the simulcast in non-Chrome browser and priority-based on Firefox.
- Fix the issue where selecting webAudio or background blur clear out the form.

**Testing**

- Join a meeting with 3 videos in Chrome with simulcast and priority-based downlink policies.
- Verify that simulcast is enabled 
- Verify that zoom in and out button only shown when priority-based downlink policy is enabled.

2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? Yes, the meeting demo.

### DIFF
--- a/apps/meeting/src/containers/MeetingForm/index.tsx
+++ b/apps/meeting/src/containers/MeetingForm/index.tsx
@@ -1,67 +1,73 @@
 // Copyright 2020-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: MIT-0
 
-import React, { useState, useContext, ChangeEvent } from 'react';
-import { useHistory } from 'react-router-dom';
+import React, {ChangeEvent, useContext, useState} from 'react';
+import {useHistory} from 'react-router-dom';
 import {
-  Input,
   Checkbox,
+  DeviceLabels,
   Flex,
-  Heading,
   FormField,
-  PrimaryButton,
-  useMeetingManager,
+  Heading,
+  Input,
   Modal,
   ModalBody,
   ModalHeader,
-  DeviceLabels,
-  Select
+  PrimaryButton,
+  Select,
+  useMeetingManager
 } from 'amazon-chime-sdk-component-library-react';
+import { DefaultBrowserBehavior } from 'amazon-chime-sdk-js';
 
-import { getErrorContext } from '../../providers/ErrorProvider';
+import {getErrorContext} from '../../providers/ErrorProvider';
 import routes from '../../constants/routes';
 import Card from '../../components/Card';
 import Spinner from '../../components/icons/Spinner';
 import DevicePermissionPrompt from '../DevicePermissionPrompt';
 import RegionSelection from './RegionSelection';
-import { fetchMeeting, createGetAttendeeCallback } from '../../utils/api';
-import { useAppState } from '../../providers/AppStateProvider';
-import { MeetingMode, BlurValues } from '../../types';
+import {createGetAttendeeCallback, fetchMeeting} from '../../utils/api';
+import {useAppState} from '../../providers/AppStateProvider';
+import {BlurValues, MeetingMode} from '../../types';
+import meetingConfig from '../../meetingConfig';
 
 const BLUR_OPTIONS = [
-  { value: BlurValues.blurDisabled, label: "Disable Blur" }, 
-  { value: BlurValues.blur10Percent, label: "Blur CPU 10%" }, 
-  { value: BlurValues.blur20Percent, label: "Blur CPU 20%" }, 
+  { value: BlurValues.blurDisabled, label: "Disable Blur" },
+  { value: BlurValues.blur10Percent, label: "Blur CPU 10%" },
+  { value: BlurValues.blur20Percent, label: "Blur CPU 20%" },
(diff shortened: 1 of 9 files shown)


apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -1,67 +1,74 @@
...
     isWebAudioEnabled,
     blurOption: blurValue,
     setBlurValue,
     toggleWebAudio,
-    setMeetingMode
+    toggleSimulcast,
+    togglePriorityBasedPolicy,
+    setMeetingMode,
+    setMeetingId,
+    setLocalName,
```
devalevenkatesh, line 58: nit: setLocalName -> setLocalUserName from React state naming convention.

apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -179,6 +191,26 @@ const MeetingForm: React.FC = () => {
         value={blurValue}
         label="Blur"
       />
+      {/* <Video uplink and downlink policies/> */}
```
devalevenkatesh, line 194: Can we remove angle brackets for <Video uplink and downlink policies/>? Looks like a component.

apps/meeting/src/providers/VideoTileGridProvider/Utils.ts
```diff
@@ -96,8 +101,13 @@ export const calculateVideoSourcesToBeRendered = (
 export const updateDownlinkPreferences = (
   gridState: GridState,
   videoSourceState: VideoSourceState,
-  attendeeStates: { [attendeeId: string]: AttendeeState }
+  attendeeStates: { [attendeeId: string]: AttendeeState },
+  priorityBasedPolicy: VideoPriorityBasedPolicy | undefined
```
devalevenkatesh, line 105: I do not have a solution but is there a better place for this policy injection? Could you explain why do we need the `priorityBasedPolicy` in `VideoTileGridProvider`? May be @xuesichao can help here :)

apps/meeting/src/providers/VideoTileGridProvider/index.tsx
```diff
@@ -108,7 +108,10 @@ const VideoTileGridProvider: React.FC = ({ children }) => {
     };
 
     priorityBasedPolicy.addObserver(observer);
-
+    dispatch( {
```
devalevenkatesh, line 111: In FF, the priority policy will be undefined so should we dispatch only if valid here?

apps/meeting/src/providers/VideoTileGridProvider/Utils.ts
xuesichao: I don't think so, because this `Utils.ts` is actually a static file. And `updateDownlinkPreferences` method need this `VideoPriorityBasedPolicy`.

We may find some better solution, but that needs some refactoring.

apps/meeting/src/containers/MeetingForm/index.tsx
ltrung: I just follow the BlurSelection stuff below. Thought it was weird but hey consistent eh :D

apps/meeting/src/providers/VideoTileGridProvider/Utils.ts
ltrung: Well basically like Sichao said Util.js is not a component and this is needed so this is the only way unless you want me to rewrite this whole logic between video tile grid provider and state reducer :D

apps/meeting/src/providers/VideoTileGridProvider/index.tsx
ltrung: well its a demo so don't think we need to do validation here. The validation if needed should be included in the library.

xuesichao: I haven't tried the demo yet, but will this impact the QA testing?

ltrung: Not sure what you mean, they can toggle only choose this when testing in non-FF browser right? (this is done as requested from QA)

xuesichao: oh right, I forget that part. 😝

apps/meeting/src/containers/MeetingForm/index.tsx
ltrung: Whats naming convention? The setLocalName is an existing function so I did not change it.

xuesichao (dismissed):
Looks good to me, thanks for making this change!

apps/meeting/src/providers/VideoTileGridProvider/Utils.ts
devalevenkatesh: Yeah, can fix later, no worries.

apps/meeting/src/providers/VideoTileGridProvider/index.tsx
devalevenkatesh: Yeah just want to check if this is to be looked upon if not that important should be fine.

devalevenkatesh (approved):

PR #117, opened 2021-12-28
[meeting demo] Upgrade eslint and fix security alerts

devalevenkatesh:
**Issue #:**
- Get react meeting demo security vulnerability alerts to 0.

**Description of changes:**
1. Upgraded ES Lint version and fixed all warnings and errors for the added rules.
2. Removed unused packages.
3. Upgrades `react-dev-utils` to latest major version. This does not affect
us as we use a plugin which does not have any breaking changes.
4. Now `npm run build` will also lint the changes.

**Testing**

- Ran locally and did sanity checks using 2 attendees.
- Deployed the demo app serverlessly and again did the sanity checks.
- Everything worked fine as expected with no console level errors.

- [meeting](https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/meeting) demo.


xuesichao:
Shall we change single quote to double quote? Maybe we should still use singe quote rule to keep consistent.

devalevenkatesh:
> Shall we change single quote to double quote? Maybe we should still use singe quote rule to keep consistent.

I knew I would get this comment lol, sorry dint fixed in first and thanks for the review! I have fixed it.

apps/meeting/package.json
```diff
@@ -30,20 +29,14 @@
     "@types/react-router-dom": "^5.1.3",
     "@types/styled-system": "^5.1.9",
     "@types/throttle-debounce": "^2.1.0",
-    "concurrently": "5.2.0",
+    "@typescript-eslint/eslint-plugin": "^5.8.1",
+    "@typescript-eslint/parser": "^5.8.1",
     "css-loader": "^6.2.0",
-    "eslint": "^5.16.0",
-    "eslint-config-airbnb": "^17.1.1",
```
ltrung, line 36: Is there any reason we don't want to use this?

apps/meeting/src/constants/index.ts
```diff
@@ -51,4 +51,4 @@ export const SDK_LOG_LEVELS = {
   'warn': LogLevel.WARN,
   'error': LogLevel.ERROR,
   'off': LogLevel.OFF,
-}
\ No newline at end of file
+};
```
xuesichao, line 54: nit: newline end of file.

apps/meeting/src/containers/EndMeetingControl/index.tsx
```diff
@@ -54,18 +54,20 @@ const EndMeetingControl: React.FC = () => {
...
               />,
               <ModalButton
+                key="leave-meeting"
                 onClick={leaveMeeting}
                 variant="primary"
                 label="Leave Meeting"
                 closesModal
               />,
-              <ModalButton variant="secondary" label="Cancel" closesModal />
+              <ModalButton key="cancel-meeting-ending" variant="secondary" label="Cancel" closesModal />
```
xuesichao, line 70: Need to change double quote to single quote

apps/meeting/package.json
```diff
@@ -6,11 +6,10 @@
     "@types/react-dom": "^16.9.5",
     "@types/styled-components": "^4.4.2",
     "@types/webpack-env": "^1.15.1",
-    "@typescript-eslint/eslint-plugin": "^2.19.2",
-    "@typescript-eslint/parser": "^2.19.2",
     "amazon-chime-sdk-component-library-react": "^2.12.0",
     "amazon-chime-sdk-js": "^2.24.0",
     "aws-sdk": "^2.731.0",
+    "concurrently": "^6.5.1",
```
michhyun1, line 12: should this be a devdependency?

apps/meeting/src/containers/MeetingRoster.tsx
```diff
@@ -6,7 +6,8 @@ import {
   Roster,
   RosterHeader,
   RosterGroup,
-  useRosterState
+  useRosterState,
+  RosterAttendeeType
```
xuesichao, line 10: nit: add trailing comma here

apps/meeting/package.json
devalevenkatesh: I did not change any existing dependency placing, I think I only changed for the linting as I added it as a new dependency.

xuesichao (commented):
We still have many `any` type in our code, instead of disable eslint for those lines, I would prefer to disable no-explicit-any rule in our config.
```
rules: {
    "@typescript-eslint/no-explicit-any": "off"
  },
```

Maybe we should also disable `@typescript-eslint/no-empty-function` rule.

apps/meeting/package.json
devalevenkatesh: I configured the eslint using command line where it asks for whether you want to use any preferred eslint configs like if to follow airbnb and so on. I checked the last deleted eslint config file and did not find any reference to airbnb rules hence removed it.

We were also not running lint so Instead of choosing the set solution, I used our own with basic set of rules.  With this setup I fixed around (12 errors and 34 warnings). I think we can make them more stricter as we want later. Still should I try airbnb config?

apps/meeting/src/containers/MeetingRoster.tsx
devalevenkatesh: I think this eslint config removes the trailing comma. I can look on adding that rule.

apps/meeting/src/containers/EndMeetingControl/index.tsx
devalevenkatesh: HTML or JSX props are still preferred to be double quotes over single.

ltrung:
> We still have many `any` type in our code, instead of disable eslint for those lines, I would prefer to `disable no-explicit-any` rule in our config.
> 
> ```
> rules: {
>     "@typescript-eslint/no-explicit-any": "off"
>   },
> ```
> 
> Maybe we should also disable `@typescript-eslint/no-empty-function` rule.

We should enforce it and remove as we move along. Disable it entirely just open for more `any` type :)

apps/meeting/src/containers/MeetingRoster.tsx
ltrung: What is the benefit of having trailing comma?

apps/meeting/src/constants/index.ts
devalevenkatesh: > We still have many `any` type in our code, instead of disable eslint for those lines, I would prefer to `disable no-explicit-any` rule in our config.
> 
> ```
> rules: {
>     "@typescript-eslint/no-explicit-any": "off"
>   },
> ```
> 
> Maybe we should also disable `@typescript-eslint/no-empty-function` rule.

I think we should not centralize `no-explicit-any` as it is a bad practice to disable it completely rather the `any` should be avoided if we can in henceforth code changes.

A more better rule would be allow any for the comments describing why it is used and that can be a centralized rule. I did not have time to go over each of them to fix so disabled on usage as of now.

apps/meeting/src/containers/EndMeetingControl/index.tsx
ltrung: For giggle https://github.com/airbnb/javascript/issues/629 :D

apps/meeting/package.json
ltrung: no just want to check whether you think we should follow airbnb as it is quite popular. Anyway, we can always revisit this when we have a single centralize eslint config :)

devalevenkatesh: Let me do a quick check with airbnb setup and get back if I can fix all lets use it as a standard solution for now, if overwhelming errors and warning I will keep this setup for now 😅

apps/meeting/src/containers/EndMeetingControl/index.tsx
devalevenkatesh: Lol, there should be JSX rules to auto-fix this I think. But since we are using double quotes for the props from start lets keep it that way.

apps/meeting/src/containers/MeetingRoster.tsx
xuesichao: One of the benefit is one you try to add/remove one line, the git diff is one-line change, instead of two.
For example:
```
// No trailing comma
{
  RosterGroup,
  useRosterState,
  RosterAttendeeType
}

// A new change requires two-line change
{
  RosterGroup,
  useRosterState,
  RosterAttendeeType,    // <=== add comma here
  NewDependency  // <==== add new line here
}
```

This makes version control diffs cleaner and editing code might be less troublesome. 

Just a best practice like new line end of file.

ltrung: i see. Make sense.  I never remember it. Thank @xuesichao

apps/meeting/package.json
devalevenkatesh: So, I got below errors with airbnb style guide:
✖ 447 problems (437 errors, 10 warnings)

We can re-visit this later 😅 .

apps/meeting/package.json
devalevenkatesh: Moved this to dev dependency.

apps/meeting/src/containers/MeetingRoster.tsx
devalevenkatesh: I have updated the code base and added trailing commas rule.

apps/meeting/.eslintrc.json
```diff
@@ -0,0 +1,56 @@
+{
+    "env": {
+        "browser": true,
+        "es2021": true
+    },
```
xuesichao, line 5: The indent of this file is 4 spaces, do we need to update it?

apps/meeting/.eslintrc.json
```diff
@@ -0,0 +1,56 @@
...
+        ],
+        "comma-dangle": [
+            "error",
+            {
+                "arrays": "always",
+                "objects": "always",
+                "imports": "always",
+                "exports": "always"
+            }
+        ],
```
xuesichao, line 49: I'm not sure about it, but adding trailing comma in the one-line statement looks weird to me.
```js
import { useNavigation, } from '../../providers/NavigationProvider';
```
Instead of "always" I would prefer "only-multiline". Do you think it's better to update it to below?
```
"comma-dangle": ["error", "only-multiline"],
```

devalevenkatesh: Used "always-multiline" over "only-multiline" to enforce the rule with "only-multiline" it is allowed and not required but with "always" it is required.

```
// "always-multiline"
var foo = {
    bar: "baz",
    qux: "quux", <- Required
};

// "only-multiline"
var foo = {
    bar: "baz",
    qux: "quux" <- Is ok to not add a comma here.
};
```

Source: https://eslint.org/docs/rules/comma-dangle

apps/meeting/.eslintrc.json
devalevenkatesh: Updated to 2 space indentation.

xuesichao: It still looks like 4 indent, is this GitHub display issue?
image

xuesichao (approved):
LGTM, but you may want to get another approval.
Thanks for upgrading these. 👍

ltrung (approved):

PR #120, opened 2022-01-05
Fixed Web Audio not being enabled properly, disabled the ability to h…

LeviSklaroff:
…ave web audio and disable audio video enabled at once because of a bug and redundant logic

**Issue #:**
There's a bug that prevents Web Audio from being enabled properly, there is also a bug where if a user with WebAudio and Audio and Video disable join start a meeting they will instantly get kicked.
**Description of changes:**
Fixed the bug where Web Audio enabled would not be passed correctly. Prevent the ability to enable both web audio and disable audio video.
**Testing**

- Test to confirm you can not enable both web audio and disable audio video.
- Join a meeting as two users with web audio
- Enable voice focus
- Test for voice focus
Meeting demo

### DIFF
--- a/apps/meeting/src/containers/MeetingForm/index.tsx
+++ b/apps/meeting/src/containers/MeetingForm/index.tsx
@@ -96,6 +96,7 @@ const MeetingForm: React.FC = () => {
             : DeviceLabels.AudioAndVideo,
         meetingManagerConfig: {
           ...meetingConfig,
+          enableWebAudio: isWebAudioEnabled,
           simulcastEnabled: enableSimulcast,
         },
       });


apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -169,6 +171,7 @@ const MeetingForm: React.FC = () => {
             setMeetingMode(MeetingMode.Attendee);
           } else {
             setMeetingMode(MeetingMode.Spectator);
+            setIsWebAudioEnabled(false);
```
michhyun1, line 174: can you jsut check if isWebAudioEnabled first and then toggleWebAudio if needed

LeviSklaroff: Yep, it reduces the changes, so I went ahead a changed it to that.

apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -177,7 +180,10 @@ const MeetingForm: React.FC = () => {
         label="Enable Web Audio"
         value=""
         checked={isWebAudioEnabled}
-        onChange={toggleWebAudio}
+        onChange={(): void => {
+          toggleWebAudio();
+          setMeetingMode(MeetingMode.Attendee);
```
ltrung, line 185: Why doesn't WebAudio not work with spectator mode?

michhyun1: Basically some weird error because we havent called getUserMedia at all, the attendee is somehow dropped from the meeting. So we thought it doesnt make sense to have both of these checkboxes enabled at the same time anyways so just disable one when the other is checked.

ltrung: Can we reproduce the issue in JS SDK? Feel like we need to know why the attendee getting dropped.

LeviSklaroff: Hard to say for sure, I haven't found a good way to reproduce the issue the SDK. I was going to create a sim to track separately to root cause better and depending on the root cause generate a warning.

ltrung: So is there a need to fix this right now before we root cause it?

michhyun1: Technically this isnt release-blocking right? If so, then maybe its not needed right now actually

LeviSklaroff: We could leave it for now, but it's not very good behavior and would be good to at least put in a temporary fix. If the user joins the meeting with WebAudio which is on by default and Disable Audio Video, they don't see the roster and can't hear the meeting.

ltrung: it is not release blocking so I would vote for root cause it first :)

LeviSklaroff: Ok, I'll revert this fix and create a sim to track

LeviSklaroff: Updated, now the fix just enables webaudio properly

ltrung (approved):

PR #150, opened 2022-02-10
Typing Indicator demo changes

eliasdem:
**Issue #:**

**Description of changes:**
Typing Indicator demo
     * Add typing indicator for channels
     * minor fix related to presence demo to publish offline status when user logs out

**Testing**

tested the feature with existing chat sdk demo app features.
yes. chat sdk demo app (channel flows demo, meetings integration demo)

### DIFF
--- a/apps/chat/src/containers/TypingIndicator/index.tsx
+++ b/apps/chat/src/containers/TypingIndicator/index.tsx
@@ -1,0 +1,55 @@
+// Copyright 2020-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT-0
+
+import React, { useEffect } from 'react';
+import styled from 'styled-components';
+import { ChatBubble, } from "amazon-chime-sdk-component-library-react";
+
+import { useChatChannelState, } from "../../providers/ChatMessagesProvider";
+
+import "./index.css";
+
+const TypingIndicatorSection = styled.div`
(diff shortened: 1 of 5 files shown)


apps/chat/src/containers/TypingIndicator/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const {
+    typingIndicator,
+    setTypingIndicator,
+  } = useChatChannelState();
+
+  useEffect(() => {
+    if (typingIndicator) {
+      (async function x() {
+        setTimeout(() => {
+          setTypingIndicator(null);
```
LeviSklaroff, line 27: Have you tested if a user is typing for more than 30 seconds, does this temporarily cause it to appear as if they're not typing?

apps/chat/src/containers/input/Input.jsx
```diff
@@ -33,7 +36,6 @@ const uploadObjDefaults = {
   key: '',
 };
 
-import { useAuthContext } from '../../providers/AuthProvider';
```
LeviSklaroff, line 39: nit. spacing

apps/chat/src/containers/TypingIndicator/index.tsx
eliasdem: no, just tested it and the indicator is shown as expected.

LeviSklaroff (approved):

PR #164, opened 2022-03-16
Add amazon-chime-sdk-component-library-react submodule

ltrung:
**Description of changes:**
Add `amazon-chime-sdk-component-library-react` submodule so that the [meeting demo](https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/meeting), which is often used to test development changes in `amazon-chime-sdk-component-library-react`, can use the latest changes from the library repo without waiting for npm release.

**Testing**

1. How did you test these changes? Manually run npm command.
2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? No

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "amazon-chime-sdk-component-library-react"]
+	path = amazon-chime-sdk-component-library-react
+	url = git@github.com:aws/amazon-chime-sdk-component-library-react.git
--- a/apps/meeting/package.json
+++ b/apps/meeting/package.json
@@ -6,8 +6,8 @@
     "@types/react-dom": "^16.9.5",
     "@types/styled-components": "^4.4.2",
     "@types/webpack-env": "^1.15.1",
-    "amazon-chime-sdk-component-library-react": "^2.15.0",
-    "amazon-chime-sdk-js": "^2.26.0",
+    "amazon-chime-sdk-component-library-react": "file:../../amazon-chime-sdk-component-library-react",
+    "amazon-chime-sdk-js": "^3.0.0-beta.1",
     "aws-sdk": "^2.1034.0",
     "fs-extra": "^10.0.0",
     "lodash.isequal": "^4.5.0",
@@ -20,10 +20,13 @@
   },
   "scripts": {
     "lint": "eslint src --ext .ts,.tsx --fix",
-    "build": "npm run lint && webpack --config ./webpack.config.js",
+    "preinstall": "scripts/setup-chime-react-submodule.js",
+    "build:fast": "webpack --config ./webpack.config.js",
+    "build": "npm run lint && npm install && npm run build:fast",
     "start:client": "webpack serve --config ./webpack.config.dev.js",
     "start:backend": "node server.js",
-    "start": "concurrently \"npm run start:client\" \"npm run start:backend\""
+    "start:fast": "concurrently \"npm run start:client\" \"npm run start:backend\"",
+    "start": "npm install && npm run start:fast"
   },
   "devDependencies": {
     "@types/lodash.isequal": "^4.5.5",
--- a/apps/meeting/package-lock.json
+++ b/apps/meeting/package-lock.json
@@ -7,13 +7,14 @@
     "": {
       "name": "chime-sdk-meeting-demo",
       "version": "0.1.0",
+      "hasInstallScript": true,
       "dependencies": {
         "@types/react": "^16.9.46",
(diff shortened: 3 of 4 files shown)


apps/meeting/scripts/setup_chime_react_submodule.js
```diff
@@ -0,0 +1,42 @@
...
+    console.log(`Command ${command} failed with exit code ${cmd.status} signal ${cmd.signal}`);
+    console.log(cmd.stderr.toString());
+    process.exit(cmd.status);
+  }
+  return output;
+}
+
+process.chdir(path.join(__dirname, '../../../amazon-chime-sdk-component-library-react'));
+
+if (!fs.existsSync('node_modules')) {
```
devalevenkatesh, line 32: Should we also handle, if the node_modules exists and there are changes in package.json then we need to re-install?

apps/meeting/scripts/setup_chime_react_submodule.js
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env node
```
devalevenkatesh, line 1: In JS SDK [demo/scripts](https://github.com/aws/amazon-chime-sdk-js/tree/main/demos/browser/script) we follow a dashed naming convention, can we follow the same here: `setup-chime-react-submodule.js`?

apps/meeting/scripts/setup_chime_react_submodule.js
devalevenkatesh: May be do this everytime? Or have to add details to do so if any changes in react component library.

apps/meeting/package.json
```diff
@@ -20,10 +20,13 @@
...
     "lint": "eslint src --ext .ts,.tsx --fix",
-    "build": "npm run lint && webpack --config ./webpack.config.js",
+    "preinstall": "scripts/setup_chime_react_submodule.js",
+    "build:fast": "webpack --config ./webpack.config.js",
+    "build": "npm run lint && npm install && npm run build:fast",
     "start:client": "webpack serve --config ./webpack.config.dev.js",
     "start:backend": "node server.js",
-    "start": "concurrently \"npm run start:client\" \"npm run start:backend\""
+    "start:fast": "concurrently \"npm run start:client\" \"npm run start:backend\"",
+    "start": "npm install && npm run start:fast"
```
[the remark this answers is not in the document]
simmkyu: Yeah, I don't think running `npm install`, especially in the `start` script, is a good pattern.

apps/meeting/scripts/setup_chime_react_submodule.js
ltrung: Why do we need to reinstall? This is just for the submodule.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "amazon-chime-sdk-component-library-react"]
+	path = amazon-chime-sdk-component-library-react
+	url = git@github.com:aws/amazon-chime-sdk-component-library-react.git
```
xuesichao, line 3: The indention here seems to be not correct, could you please verify it?

ltrung: This is auto-generated by git (i.e., I didn't touch it) :D

apps/meeting/scripts/setup_chime_react_submodule.js
devalevenkatesh: So you are changing into the `amazon-chime-component-library..` and then you are checking for `node_module` if not existing and then only run the rest commands, hence, what would be the case:
1. To update the component library to a newer commit?
2. If the newer commit contains a new dependency in component library and locally I have the `node_modules` already it wont update right?

ltrung: You have to manually do it in that case anyway. This is purely a setup step to run it. I think it is the same way in JS SDK demo (we don't run npm install every time).

apps/meeting/scripts/setup_chime_react_submodule.js
ltrung: Fixed.

apps/meeting/scripts/setup_chime_react_submodule.js
devalevenkatesh: Ok, yeah I think my concern is just to let builders know how to update it :)

devalevenkatesh (approved):

xuesichao (approved):
Looks good to me.

xuesichao (approved):

PR #181, opened 2022-04-04
Add chat components based on DataMessages

devalevenkatesh:
**Issue #:**

**Description of changes:**
- Add DataMessagesProvider to manage the state of messages.
- Add a Messages component to use the messages from DataMessagesProvider
  and show to UI.
- Add demo based light and dark theme to accomodate theme for Chat components.
  This theme is extension of existing component library's themes + chat specific
  theme.
- Adjust Navigation components to support Chat components with responsive UI
  similar to Roster component.

**Testing**

- Test manually joining same meeting with different attendees.
- Checked for error or warning in console logs.
- Checked message sending and receiving works as expected.
- Tested responsive UI, check screenshots.

Screen Shot 2022-04-03 at 11 14 55 PM
Screen Shot 2022-04-03 at 11 15 07 PM
Screen Shot 2022-04-03 at 11 15 23 PM
Screen Shot 2022-04-03 at 11 15 31 PM

apps/meeting

### DIFF
--- a/apps/meeting/src/containers/Chat/Messages.tsx
+++ b/apps/meeting/src/containers/Chat/Messages.tsx
@@ -1,0 +1,33 @@
+import React, { useEffect, useRef } from 'react';
+import { ChatBubble } from 'amazon-chime-sdk-component-library-react';
+import { useDataMessages } from '../../providers/DataMessagesProvider';
+import { StyledMessages } from './Styled';
+
+export default function Messages() {
+  const { messages } = useDataMessages();
+  const scrollRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (messages.length === 0) {
+      return;
+    }
+    if (scrollRef.current) {
+      scrollRef.current.scrollTop = scrollRef.current.scrollHeight;
+    }
+  }, [messages.length]);
+
+  const renderMessages = () => {
+    return messages.map((message) => (
+      <ChatBubble
+        variant={message.isSelf ? 'outgoing' : 'incoming'}
+        senderName={message.senderName}
+        key={message.timestamp}
+        marginLeft="1rem"
+      >
+        {message.message}
+      </ChatBubble>
+    ));
+  };
+
+  return <StyledMessages ref={scrollRef}>{renderMessages()}</StyledMessages>;
+}
--- a/apps/meeting/src/providers/DataMessagesProvider/index.tsx
+++ b/apps/meeting/src/providers/DataMessagesProvider/index.tsx
@@ -1,0 +1,109 @@
+import { RosterType, useAudioVideo, useMeetingManager, useRosterState } from 'amazon-chime-sdk-component-library-react';
+import { DataMessage } from 'amazon-chime-sdk-js';
+import React, { useEffect, useReducer, createContext, useContext, FC, useCallback, useRef } from 'react';
+import { DataMessagesActionType, initialState, ChatDataMessage, reducer } from './state';
+
+interface DataMessagesStateContextType {
+  sendMessage: (message: string) => void;
+  messages: ChatDataMessage[];
+}
+
+const DataMessagesStateContext = createContext<DataMessagesStateContextType | undefined>(undefined);
+
+export const DataMessagesProvider: FC = ({ children }) => {
+  const meetingManager = useMeetingManager();
+  const audioVideo = useAudioVideo();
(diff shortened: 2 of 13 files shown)


apps/meeting/src/containers/Chat/Messages.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useRef } from 'react';
+import { ChatBubble } from 'amazon-chime-sdk-component-library-react';
+
```
xuesichao, line 3: nit: remove this empty space

apps/meeting/src/providers/DataMessagesProvider/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const handler = useCallback(
+    (dataMessage: DataMessage) => {
+      if (!dataMessage.throttled) {
+        const isSelf =
+          dataMessage.senderAttendeeId === meetingManager.meetingSession?.configuration.credentials?.attendeeId;
+        if (isSelf) {
+          dispatch({
+            type: DataMessagesActionType.ADD,
+            payload: {
+              message: new TextDecoder().decode(dataMessage.data),
```
xuesichao, line 45: For understanding, why we need to handle `dataMessage.data` differently than line 57.

```js
message: new TextDecoder().decode(dataMessage.data),
message: data.message,
```

apps/meeting/src/providers/DataMessagesProvider/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+            type: DataMessagesActionType.ADD,
+            payload: {
+              message: data.message,
+              senderAttendeeId: dataMessage.senderAttendeeId,
+              timestamp: dataMessage.timestampMs,
+              senderName: rosterRef.current[dataMessage.senderAttendeeId]?.name!,
+              isSelf: false,
+            },
+          });
+        }
```
xuesichao, line 64: I think we can merge these two conditions.
```js
dispatch({
  type: DataMessagesActionType.ADD,
  payload: {
    message: isSelf? new TextDecoder().decode(dataMessage.data): data.message, // <====
    senderAttendeeId: dataMessage.senderAttendeeId,
    timestamp: dataMessage.timestampMs,
    senderName: rosterRef.current[dataMessage.senderAttendeeId]?.name!,
    isSelf: isSelf, // <====
  },
});
```

xuesichao (commented):
Looks good to me in general!

apps/meeting/src/providers/DataMessagesProvider/index.tsx
devalevenkatesh: So, when a local attendee sends the data message, it is sent over the peer-connection for remotes + triggered to self so that the the subscribed data message handler gets all updates in one place. This is the same logic in JS SDK as well. Basically, update the state only when the data message has successfully received from local attendee or remote.

devalevenkatesh: For a local attendee, the data message received from the remote is a JSON, but when sending from local to remote, the new DataMessage creation accepts only Uint8Array and not a string hence had to handle separate cases.

apps/meeting/src/providers/DataMessagesProvider/index.tsx
devalevenkatesh: Actually, in `else` here, I have to do `dataMessage.json()` to get the `data` which is not the case for the "self" path. The self path received DataMessage's message is a Uint8Array and not stringified JSON hence handled in separate cases.

apps/meeting/src/containers/Chat/Messages.tsx
devalevenkatesh: Fixed it, thanks!

apps/meeting/src/providers/DataMessagesProvider/index.tsx
xuesichao: Got it! Thanks for your detailed explanation. 👍
Learned something new :D.

apps/meeting/src/providers/DataMessagesProvider/index.tsx
xuesichao: Oh that's right!

xuesichao (approved):
LGTM

PR #184, opened 2022-04-05
Use logger from component library

devalevenkatesh:
**Issue #:**

**Description of changes:**
- Necessary changes to use logger from component library.
- Component library PR: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/772

**Testing**

- Deployed it and checked whether we get console logs as well as logs in AWS CloudWatch in the deployed account.
- If `LoggerProvider` is not mounted, only console logger will work as it is provided by default.

- apps/meeting

### DIFF
--- a/apps/meeting/serverless/src/handlers.js
+++ b/apps/meeting/serverless/src/handlers.js
@@ -325,7 +325,7 @@ exports.logs = async (event, context) => {
         const timestampIso = new Date(log.timestampMs).toISOString();
         let message = `${body.appName} ${timestampIso} [${log.sequenceNumber}] [${log.logLevel}]`;
         if (body.meetingId && body.attendeeId) {
-          message = `${message} [mid: ${body.meetingId.toString()}] [aid: ${body.attendeeId}]: ${log.message}`;
+          message = `${message} [meetingId: ${body.meetingId.toString()}] [attendeeId: ${body.attendeeId}]: ${log.message}`;
         } else {
           message = `${message}: ${log.message}`;
         }
--- a/apps/meeting/serverless/src/handlers.js
+++ b/apps/meeting/serverless/src/handlers.js
@@ -293,37 +293,42 @@ async function ensureLogStream(cloudWatchClient, logStreamName) {
 }
 
 exports.logs = async (event, context) => {
-  var response = {
+  const response = {
     "statusCode": 200,
     "headers": {},
     "body": '',
     "isBase64Encoded": false
   };
   {
-    var body = JSON.parse(event.body);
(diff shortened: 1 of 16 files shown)


apps/meeting/serverless/src/handlers.js
```diff
@@ -293,37 +293,42 @@ async function ensureLogStream(cloudWatchClient, logStreamName) {
...
     if (body.logs.length > 0) {
       for (let i = 0; i < body.logs.length; i++) {
-        var log = body.logs[i];
-        var timestampIso = new Date(log.timestampMs).toISOString();
-        var message = `${timestampIso} [${log.sequenceNumber}] [${log.logLevel}] [mid: ${body.meetingId.toString()}] [aid: ${body.attendeeId}]: ${log.message}`;
+        const log = body.logs[i];
+        const timestampIso = new Date(log.timestampMs).toISOString();
+        let message = `${body.appName} ${timestampIso} [${log.sequenceNumber}] [${log.logLevel}]`;
+        if (body.meetingId && body.attendeeId) {
+          message = `${message} [mid: ${body.meetingId.toString()}] [aid: ${body.attendeeId}]: ${log.message}`;
```
ltrung, line 328: nit: Can we spell out meetingId and attendeeId? More space but it is clearer. mid is a shorthand for media Id in webRTC and aid is an English word :)

apps/meeting/src/meetingConfig.ts
```diff
@@ -1,38 +1,32 @@
...
-  postLogConfig,
-  logger,
-};
+if (!['0.0.0.0', '127.0.0.1', 'localhost'].includes(location.hostname)) {
+  const postLogger = new POSTLogger({
+    url: `${BASE_URL}logs`,
+    logLevel,
+    metadata: {
+      appName: 'ChimeComponentLibraryReactDemo',
+      timestamp: Date.now().toString(), // Add current timestamp for unique AWS CloudWatch log stream generation. This will be unique per POSTLogger creation in time.
```
ltrung, line 25: unlikely but it is not unique if 2 ppl join at the exact same time right? Though our demo does not handle that case either so its probably fine but feel like just use uuid here is better?

Also could we use external meetingId and attendeeId so we could find the log easier?

apps/meeting/serverless/src/handlers.js
devalevenkatesh: Yup makes sense updated!

apps/meeting/src/meetingConfig.ts
devalevenkatesh: Will have to come back to this, but will track!

PR #191, opened 2022-04-19
Fix the issue that background feature enables camera unexpectedly

xuesichao:
**Issue #:**
1. `VideoInputTransformControl` starts the camera unexpectedly when calling `resetDeviceToIntrinsic` after mounting.
2. When local video is not enabled, enabling the background feature via `VideoInputTransformControl` drop-down menu will accidentally enable the camera.

**Description of changes:**
1. Use `selectVideoInputDevice` API instead of `startVideoInputDevice` API in `resetDeviceToIntrinsic` function.
2. Change `VideoInputTransformControl` to :
  * call `startVideoInputDevice` if the local video is already enabled
  * call `selectVideoInputDevice` if the local video is not enabled

**Testing**

Run the meeting demo and confirm it works as expected.

Yes, meeting demo

2. If applicable, have you run `npm run build` successfully locally to fix all warnings and errors?
Yes

### DIFF
--- a/apps/meeting/src/components/DeviceSelection/CameraDevices/VideoTransformDropdown.tsx
+++ b/apps/meeting/src/components/DeviceSelection/CameraDevices/VideoTransformDropdown.tsx
@@ -55,7 +55,7 @@ export const VideoTransformDropdown: React.FC<Props> = ({
   ];
 
   // Creates a device based on the selections (None, Blur, Replacement) and uses it as input.
-  async function selectTransform(e: ChangeEvent<HTMLSelectElement>) {
+  const selectTransform = async (e: ChangeEvent<HTMLSelectElement>) => {
     const selectedTransform = e.target.value;
     let currentDevice = selectedDevice;
 
@@ -86,7 +86,7 @@ export const VideoTransformDropdown: React.FC<Props> = ({
     } finally {
       setIsLoading(false);
     }
-  }
+  };
 
   return (
     <FormField
--- a/apps/meeting/src/components/MeetingControls/VideoInputTransformControl.tsx
+++ b/apps/meeting/src/components/MeetingControls/VideoInputTransformControl.tsx
@@ -51,28 +51,30 @@ const VideoInputTransformControl: React.FC<Props> = ({
     resetDeviceToIntrinsic();
   }, []);
 
-  // Reset the video input to intrinsic if current video input is a transform device because this component does not know if blur or replacement was selected.
-  // This depends on how the demo is set up.
-  // TODO: use a hook in the appState to track whether blur or replacement was selected before this component mounts.
-  async function resetDeviceToIntrinsic() {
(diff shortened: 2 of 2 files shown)


apps/meeting/src/components/MeetingControls/VideoInputTransformControl.tsx
```diff
@@ -51,28 +51,30 @@ const VideoInputTransformControl: React.FC<Props> = ({
     resetDeviceToIntrinsic();
   }, []);
 
-  // Reset the video input to intrinsic if current video input is a transform device because this component does not know if blur or replacement was selected.
-  // This depends on how the demo is set up.
-  // TODO: use a hook in the appState to track whether blur or replacement was selected before this component mounts.
-  async function resetDeviceToIntrinsic() {
+  // Reset the video input to intrinsic if current video input is a transform device because this component
+  // does not know if blur or replacement was selected. This depends on how the demo is set up.
+  // TODO: use a hook in the appState to track whether blur or replacement was selected before this component mounts,
```
nainkunal933, line 56: Why is this TODO there? Is it there to remove the `isVideoTransformDevice` call in the future?

xuesichao: So we have a `VideoTransformDropdown` component in the device setup page. We can preview the camera with those background features. If we select a background feature there, the `selectedVideoInputDevice` in `MeetingManager` will be updated. However the operation of enabling background feature is not tracked anywhere. So after we enter the meeting page. Although we selected a background transform device (background feature is enabled), but for the other components, they don't know that the background feature is enabled. So this TODO is mostly to fix this issue.

> Is it there to remove the isVideoTransformDevice call in the future

Currently we check if the background feature is enabled by checking the type of `selectedVideoInputDevice`, the `isVideoTransformDevice` call here. This is not very ideal from a long term. If we finish this TODO item, I think we are able to get rid of it.

nainkunal933 (approved):

apps/meeting/src/components/DeviceSelection/CameraDevices/VideoTransformDropdown.tsx
```diff
@@ -55,7 +55,7 @@ export const VideoTransformDropdown: React.FC<Props> = ({
   ];
 
   // Creates a device based on the selections (None, Blur, Replacement) and uses it as input.
-  async function selectTransform(e: ChangeEvent<HTMLSelectElement>) {
+  const selectTransform = async (e: ChangeEvent<HTMLSelectElement>) => {
```
LeviSklaroff, line 58: What's the purpose for changing each of these? Is it for styling?

xuesichao: Just for styling, no specific reason. We use arrow function most in the React SDK.

PR #220, opened 2022-08-09
Fix all security vulnerabilities in telehealth-widget demo

devalevenkatesh:
**Issue #:**
Security vulnerabilities detected.

**Description of changes:**
- `npm update --save` in all frontend, backend and lambda directories
  which hold the package.json

**Testing**

- Tested by re-deploying the app and checked if all features work as
  expected.

- apps/telehealth-widget

NA


michhyun1:
Just curious - is it better to update all deps or do npm audit fix?

michhyun1 (approved):

devalevenkatesh:
> Just curious - is it better to update all deps or do npm audit fix?

Generally we should go ahead and update the dependencies if they do not error in build process or demos working. The main reason is it this keeps the demo updated with new dependencies.

We should generally follow `npm audit fix` if `npm update` does not fix anything or runs into any errors and the fix is available through `npm audit fix`.

PR #226, opened 2022-08-18
chat demo bug fix

dpwspoon:
**Issue #:**

**Description of changes:**

* temporarily have sam directions point to run app locally since cloudfront endpoint is not working
* Fix user loging
* Partial fix for stack clean up
* Partial fix for SAM by removing global s3 bucket
* Remove generous aws role permissions

**Testing**

Run SAM steps in README then create and login as user

3. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it?

yes the chat

5. If applicable, have you run `npm run build` successfully locally to fix all warnings and errors?

yes


apps/chat/package.json
```diff
@@ -16,8 +16,9 @@
     "@aws-amplify/core": "^4.3.12",
     "@aws-amplify/storage": "^4.4.13",
     "amazon-chime-sdk-component-library-react": "^3.2.0",
+    "amazon-chime-sdk-js": "^3.6.0",
```
devalevenkatesh, line 19: Sichao had removed the JS SDK dependency as component library includes it as a dependency, any reason to add this?

dpwspoon: Needed a higher version/latest release hence why added it.  Is there a better way to do that?

devalevenkatesh: I think its fine just wanted to check cause I updated the component library to latest JS SDK but we have not released it yet.

devalevenkatesh (approved):

apps/chat/README.md
```diff
@@ -19,10 +19,13 @@ cd apps/chat
 npm run deploy
 ```
 
-3. Verify that the `src/backend/serverless/appconfig.json` is populated with the CloudFormation stack outputs. This JSON file used to
-   configure the app with `src/Config.js.`
+3. Run app locally
```
eliasdem, line 22: this section is to serve the demo app via s3 + CF. there is another section below this to run it locally.

dpwspoon: s3 + CF currently doesn't work so hence combining the two now for temporary solution

Issue #224, opened 2022-08-16
chat demo sam directions do not work for launching/components like auth not working


dpwspoon:
**Demo**
Directions here do not work: https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/chat#using-aws-sam

**Describe the bug**
The s3 bucket name is already in use.  In addition once launched (if random name provided) you can't delete the stack with out hand holding.

**To Reproduce**
1. Follow directions here: https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/chat#using-aws-sam.
2. See failures
3. Override values here: https://github.com/aws-samples/amazon-chime-sdk/blob/main/apps/chat/src/backend/serverless/samconfig.toml#L10
4. Try to delete stack, it doesn't work

**Expected behavior**
create and delete stack happens

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

dpwspoon:
The SAM change (https://github.com/aws-samples/amazon-chime-sdk/pull/200/files) also broke credential exchange service and likely other APIs as browsers don't allow authenticated calls (at least on localhost) to * anymore:

dpwspoon:
some fixs are in https://github.com/aws-samples/amazon-chime-sdk/pull/226

PR #229, opened 2022-08-19
Restrict CORS to cloudfront and localhost origins; upgrade nodejs to 14.x

eliasdem:
Restrict CORS to cloudfront and localhost origins; upgrade nodejs to 14.x

**Issue #:** 
https://github.com/aws-samples/amazon-chime-sdk/issues/224

**Description of changes:**

Restrict CORS to CloudFront and localhost origins; 
- This addresses the sign in with credential exchange service issue mentioned in https://github.com/aws-samples/amazon-chime-sdk/issues/224

Upgrade nodejs to 14.x for the lambdas
- this is to match the min required nodejs version (which is 14.x) described in the readme.
- also the new preflight request lambda code requires `nodejs 14.x`

Update ReadMe accordingly

**Testing**

Tested the chat demo app with the following steps:
- Run `npm run deploy` to deploy the demo app backend and frontend resources
- Open the CloudFront endpoint in a browser
- Sign in with `Credential Exchange Service` option and it succeeds. Able to create a channel
- Sign in with Cognito also works
- Create meetings in a channel also works

3. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? Yes, the chat demo app

5. If applicable, have you run `npm run build` successfully locally to fix all warnings and errors? yes

### DIFF
--- a/apps/chat/src/backend/serverless/template.yaml
+++ b/apps/chat/src/backend/serverless/template.yaml
@@ -1,5 +1,3 @@
-#NOTE: In a production environment, we recommend that you use a specific origin for the `Access-Control-Allow-Origin` response header
-#to scope-down access to resources.
 AWSTemplateFormatVersion: '2010-09-09'
 Transform: "AWS::Serverless-2016-10-31"
 Description: AWS CloudFormation stack to set up infrastructure required for the Amazon Chime SDK chat demo app
@@ -42,7 +40,7 @@ Resources:
     Description: The AWS SDK with support for Amazon Chime SDK messaging features.
     Properties:
       CompatibleRuntimes:
-        - "nodejs12.x"
+        - nodejs14.x
       Content:
         S3Bucket: aws-blog-business-productivity-chime-sdk
         S3Key: chat-sdk-demo/aws-sdk-chime-layer.zip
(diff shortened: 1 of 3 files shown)


apps/chat/src/backend/serverless/template.yaml
```diff
@@ -300,7 +298,7 @@ Resources:
     Properties:
       Handler: "index.handler"
       Role: !GetAtt LambdaExecuteRole.Arn
-      Runtime: "nodejs12.x"
+      Runtime: nodejs14.x
```
ltrung, line 301: Could we do node 18 now? 14 is on the way out already :)

devalevenkatesh: Lambda does not support 18 but 16 is supported I think.

ltrung:
For demo, do we even need to enable CORS?

eliasdem:
> For demo, do we even need to enable CORS?

yea, normally browsers enforce CORS checks unless you disable the security features. Also using `*` doesn't work with credential exchange in API Gateway (the bug reported in the issue linked above)

ltrung (approved):

PR #244, opened 2022-10-25
Updates to cloudfront settings: now available globally and being able to support retries

gmiti-amazon:
**Description of changes:**
- Fix for s3 bucket name for chat demo when using attachments
- Updates to cloudfront settings: now available globally and being able to support retries

**Testing**

1. How did you test these changes? - Ran the sam deploy scripts, and the application is deploying properly.
2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? chat
N/A


apps/chat/src/backend/serverless/template.yaml
```diff
@@ -1599,12 +1599,16 @@ Resources:
...
-              - US
-              - JP
-              - AU
-              - FR
+            RestrictionType: none
+        CustomErrorResponses:
+          - ErrorCachingMinTTL: 300
+            ErrorCode: 404
+            ResponseCode: 200
+            ResponsePagePath: /chat.html
```
devalevenkatesh, line 1607: So here for 404, 403s we show the chat.html page. What does it look like and is that a static show page? Having a screenshot in description would help :)

Also, description mentions a S3 bucket fix and I also see the commit but no file change for it in commit?

PR #298, opened 2023-04-05
add code samples for call insights

lognoel:
Adds sample code for call analytics data lake


lambda/call-insights-parsing/docker/app.py
```diff
@@ -0,0 +1,137 @@
+import json
```
ltrung, line 1: Just curious. Any reason to use python over node? Most of our team are more familiar with Node :D

ltrung (approved):

PR #306, opened 2023-04-10
Fix attachment bug when cancel the attachment selection

jingwy:
**Issue #:** 306

**Description of changes:** 

Fix attachment bug in chat and televisit demo app when user sends a message with attachment. The bug is, when sending a message with an attachment, after user selects the attachment, if user click on the "x" to cancel/de-select the attachment, the current bug will send the attachment instead. This code change fixes the behavior to allow user to de-select or cancel the attachment.

**Testing** 

Run the demo app locally and verified that the issue is fixed.
3. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? Yes the chat demo application can be used to test it.
4. If applicable, have you run `npm run build` successfully locally to fix all warnings and errors? Yes

### DIFF
--- a/apps/chat/src/containers/input/Input.jsx
+++ b/apps/chat/src/containers/input/Input.jsx
@@ -162,6 +162,7 @@ const Input = ({ activeChannelArn, member, hasMembership }) => {
       className="write-link attach"
       onClick={(_event) => {
         uploadRef.current.value = null;
+        setUploadObj(uploadObjDefaults);
         uploadRef.current.click();
       }}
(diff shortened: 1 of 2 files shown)


ltrung:
Do you also want to fix it in this demo https://github.com/aws-samples/amazon-chime-sdk/blob/c1916bef61478f5f374de109c7e49a89364ab203/apps/televisit-demo/frontend/src/containers/input/Input.jsx#L110 ?

ltrung (approved):

PR #315, opened 2023-05-11
Add `skipDeviceSelection` flag in meeting join view

devalevenkatesh:
**Issue #:**

**Description of changes:**

When one use `skipDeviceSelection` while joining a meeting, then, the user will have to manually select the devices either in device setup view or in meeting view screen.

**Testing**

- Load the meeting demo
- Put in meeting and attendee details
- Check "Skip device selection" option
- Verify that the "Device setup" view shows next.
- Select or change the devices manually to select the audio and video devices and then join the meeting.
- Verify in meeting the attendee is seen in roster and can start video and sees it, also the mic indicator is active when talked.
- Do the same steps without checking "Skip default meeting join device listing and selection" option to make sure both enabling/disabling "Skip default meeting join device listing and selection" option works.

2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? apps/meeting demo

Yes

### DIFF
--- a/apps/meeting/src/containers/MeetingForm/index.tsx
+++ b/apps/meeting/src/containers/MeetingForm/index.tsx
@@ -62,6 +62,8 @@ const MeetingForm: React.FC = () => {
     setLocalUserName,
     setRegion,
     setCpuUtilization,
+    skipDeviceSelection,
+    toggleMeetingJoinDeviceSelection,
   } = useAppState();
   const [meetingErr, setMeetingErr] = useState(false);
   const [nameErr, setNameErr] = useState(false);
@@ -117,6 +119,7 @@ const MeetingForm: React.FC = () => {
       const options: MeetingManagerJoinOptions = {
         deviceLabels: meetingMode === MeetingMode.Spectator ? DeviceLabels.None : DeviceLabels.AudioAndVideo,
         enableWebAudio: isWebAudioEnabled,
+        skipDeviceSelection,
       };
 
       await meetingManager.join(meetingSessionConfiguration, options);
@@ -182,7 +185,7 @@ const MeetingForm: React.FC = () => {
       <RegionSelection setRegion={setRegion} region={region} />
       <FormField
         field={Checkbox}
-        label="Join w/o Audio and Video"
+        label="Join w/o Audio and Video (spectator mode)"
         value=""
         checked={meetingMode === MeetingMode.Spectator}
         onChange={(): void => {
@@ -249,6 +252,14 @@ const MeetingForm: React.FC = () => {
         value=""
         checked={keepLastFrameWhenPaused}
         onChange={toggleKeepLastFrameWhenPaused}
+      />
+      <FormField
+        field={Checkbox}
+        label="Skip meeting join device selection"
+        value=""
+        checked={skipDeviceSelection}
+        onChange={toggleMeetingJoinDeviceSelection}
+        infoText="Please select the devices manually to successfully join a meeting"
       />
       <Flex container layout="fill-space-centered" style={{ marginTop: '2.5rem' }}>
         {isLoading ? <Spinner /> : <PrimaryButton label="Continue" onClick={handleJoinMeeting} />}
--- a/apps/meeting/package-lock.json
+++ b/apps/meeting/package-lock.json
@@ -51,10 +51,18 @@
       }
     },
     "../../amazon-chime-sdk-component-library-react": {
-      "version": "0.0.1",
+      "version": "3.7.0",
+      "license": "Apache-2.0",
       "dependencies": {
-        "@mdx-js/loader": "^1.6.22",
         "@popperjs/core": "^2.2.2",
+        "fast-memoize": "^2.5.2",
+        "lodash.isequal": "^4.5.0",
+        "react-popper": "^2.2.4",
+        "throttle-debounce": "^2.3.0",
+        "uuid": "^8.0.0"
+      },
+      "devDependencies": {
+        "@mdx-js/loader": "^1.6.22",
         "@rollup/plugin-commonjs": "^13.0.0",
         "@rollup/plugin-node-resolve": "^8.0.1",
         "@storybook/addon-a11y": "6.5.8",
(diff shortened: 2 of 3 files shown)


apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -250,6 +253,14 @@ const MeetingForm: React.FC = () => {
         checked={keepLastFrameWhenPaused}
         onChange={toggleKeepLastFrameWhenPaused}
       />
+      <FormField
+        field={Checkbox}
+        label="Skip default meeting join device listing and selection"
```
xuesichao, line 258: The application is not acting as it says here when selecting this option. Thought `MeetingManager.join()` does not list and select the devices, in the `DeviceSetup` page since it uses `MicSelection` and `CameraSelection` there's still device permission prompts and list the devices.

apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -250,6 +253,14 @@ const MeetingForm: React.FC = () => {
...
+        value=""
+        checked={skipDeviceSelection}
+        onChange={toggleMeetingJoinDeviceSelection}
+        infoText="The demo will select devices on its own before meeting join"
```
xuesichao, line 262: This is a bit confusing because I don't know "The demo will select devices on its own before meeting join" happens when I select this option or not.

Suggest removing this infoText.

apps/meeting/src/views/DeviceSetup/index.tsx
```diff
@@ -6,15 +6,22 @@ import { Heading } from 'amazon-chime-sdk-component-library-react';
...
-  <StyledLayout>
+const DeviceSetup: React.FC = () => {
+  const { skipDeviceSelection } = useAppState();
+
+  return (
+    <StyledLayout>
     <Heading tag="h1" level={3} css="align-self: flex-start">
       Device settings
     </Heading>
+    {skipDeviceSelection && <DefaultDevicesSelection />}
```
xuesichao, line 20: This flow does not make sense to me. Because when selecting "skip device selection" checkbox, users expect to handle the device selection manually. But this `DefaultDevicesSelection` select the first device for audio and video automatically. It makes this flow identical with not selecting "skip device selection".

Suggest removing this line as well as `DefaultDevicesSelection`. We can manually select devices 1. in Device Setup page via device drop down or 2. in Meeting page via Meeting control.

xuesichao, line 20: nit: indention is wrong

apps/meeting/src/containers/MeetingForm/index.tsx
devalevenkatesh: I agree, but technically we want to somehow map it to skipping `meetingManager.join` API device listing and selection. Thus, I worded it so.

xuesichao: In our meeting join flow, we have below steps that lists devices:
1. `meetingManager.join`
2. `DeviceSelections` in Device Setup page
3. `MeetingControls` in Meeting page

With or without using `skipDeviceSelection`, this demo lists devices anyway, but it does skip the device selection. We can't skip listing device from meeting demo anyway, so I think changing the label to "Skip device selection" (not mentioning listing device) makes more sense. Though from API perspective we do skip device listing in `mm.join`.

This flow is bit confusing because from application perspective, we skip device selection (select first device) and automatically select first device again. Considering we have builders refer to our demo for implementation, I think it's better to make this use case clear.

Ideally we should get rid of `DeviceSelection` and `MeetingControls` so we can actually stop device listing and selection from application level.

devalevenkatesh: So does "Skip default meeting join device selection" sounds appropriate for our use case?

apps/meeting/src/containers/MeetingForm/index.tsx
devalevenkatesh: Ok will do.

apps/meeting/src/containers/MeetingForm/index.tsx
xuesichao: I slightly prefer "Skip device selection". Imaging you are a user that does't know the details of SDK. "Skip default meeting join device selection" could be confusing:
1. what is default meeting join device selection? e.g., select most recent device, select first device, ...
2. If we skip it, will this demo select any device or not?

In this flow the result is that no device is selected. So "Skip device selection"or "join w/o selecting devices" sounds more appropriate.

I'm fine as long as we don't mention "list device", decision is up to you : ).

devalevenkatesh: I have removed the `DefaultDeviceSelection` and updated the `infoText` to select the devices manually. I believe either way its going to be implicit and builder will need to be made aware on the `skipDevcieSelection` flag which is only related to `MeetingManager.Join` API. I think it is important to call "meeting join device selection" as that is the flag's exact purpose.

apps/meeting/src/views/DeviceSetup/index.tsx
devalevenkatesh: Prettified, thanks :)

apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -250,6 +253,14 @@ const MeetingForm: React.FC = () => {
         checked={keepLastFrameWhenPaused}
         onChange={toggleKeepLastFrameWhenPaused}
       />
+      <FormField
+        field={Checkbox}
+        label="Skip meeting join device selection"
+        value=""
+        checked={skipDeviceSelection}
+        onChange={toggleMeetingJoinDeviceSelection}
+        infoText="Please select the devices manually to successfully join a meeting"
```
xuesichao, line 262: device selection is not required to be able to join a meeting. It should be spectator mode in this case.

xuesichao (approved):

xuesichao (approved):

xuesichao (approved):

apps/meeting/src/containers/MeetingForm/index.tsx
devalevenkatesh: The only reason to mention this is to showcase the `skipDeviceSelection` flag, since we already have a spectator mode, this would be good to tell "Hey if you are skipping device selection at join, then you have to manually select the devices, otherwise there will be issues/no audio or fake device selection etc". Under flags docs, we explicitly mention that it would be builders responsibility to do device selection for a working meeting.

PR #349, opened 2023-08-07
Chat demo fix

jingwy:
**Issue #:** For chat demo app, during the clean up phase, where we delete the cloudformation stack, it will fail due to custom resource error. This change will fix the issue.

**Description of changes:** There is a bug where the delete event to ChimeAppInstanceAdmin lambda is of requestType "Delete" instead of "DELETE", thus fixing the code to resolve the delete stack failure.

**Testing**

1. How did you test these changes? Deployed the stack with the changes and deleted it, no more failure.
2. Can these changes be tested using one of the demo application? If yes, which demo application can be used to test it? Yes the chat demo app.
3. If applicable, have you run `npm run build` successfully locally to fix all warnings and errors?Yes
4. 


simmkyu (approved):

michhyun1 (commented):
lgtm

PR #435, opened 2024-08-05
Add Background Replacement image options in the react demo

dingyishen-amazon:
**Issue #:**
https://app.asana.com/0/1207324935836040/1207880868572886

**Description of changes:**
* Add new API `changeBackgroundReplacementImage` to the demo.
* Add a background replacement dropdown on the video preview page. 
Screenshot 2024-07-31 at 3 05 53 PM
* Add `Select Background Replacement Filter` to the Filter selection dropdown, it contains a list of background replacement filters, users could switch the filter there. 
Screenshot 2024-07-31 at 3 05 13 PM

**Testing**

Test with the react demo app. 
There is the demo video showing the new changes: 

https://github.com/user-attachments/assets/bc96c491-8d16-4acd-8413-88c99d317d6f

Yes, apps/meeting
4. If applicable, have you run `npm run build` successfully locally to fix all warnings and errors?
Yes.


michhyun1:
Do we want to show those options if there is no background replacement processor that is active?

dingyishen-amazon:
> Do we want to show those options if there is no background replacement processor that is active?

Yes, it allows users to select the replacement image before enabling the background replacement.

apps/meeting/src/components/DeviceSelection/CameraDevices/BackgroundReplacementDropdown.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const options: ReplacementDropdownOptionType[] = [
+    {
+      label: ReplacementOptions.Blue,
+      value: isBackgroundReplacementSupported === undefined || isBackgroundReplacementSupported === false ? 'Background Replacement not supported' :ReplacementOptions.Blue,
+    },
+    {
+      label: ReplacementOptions.Beach,
+      value: isBackgroundReplacementSupported === undefined || isBackgroundReplacementSupported === false ? 'Background Replacement not supported' :ReplacementOptions.Beach,
+    },
+  ];
```
xuesichao, line 37: 1. Could we merge `ReplacementDropdownOptionType` and `ReplacementOptions ` into one type? Something like using `Record<K, V>`.
- https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

2. Could we handle isSupported and placeholder in another place? The flow could be something like this:

    1. check if isSupported
    2. if true, iterate through `options` and renders dropdown options
    3. if false, iterate through a fallback `options` that renders `Background Replacement not supported`

apps/meeting/src/components/DeviceSelection/CameraDevices/BackgroundReplacementDropdown.tsx
```diff
@@ -0,0 +1,79 @@
...
+      if (selectReplacement === ReplacementOptions.Blue && isBackgroundReplacementSupported) {
+        const blob = await createColorBlob();
+        await changeBackgroundReplacementImage(blob);
+        console.info(`Video filter changed to Relacement - Blue`);
+      } else if (selectReplacement === ReplacementOptions.Beach && isBackgroundReplacementSupported) {
+        const imageInBase64 = BackgroundImageEncoding();
+        const blob = await createImageBlob(imageInBase64);
+        await changeBackgroundReplacementImage(blob);
+        console.info(`Video filter changed to Relacement - Beach`);
+      }
```
xuesichao, line 58: Could we get rid of these if else conditions and simply handles the logic based on an interface such as `options: ReplacementDropdownOptionType[]`. So we don't need to change this code to add any new options, we just need to expand the `options`

apps/meeting/src/components/DeviceSelection/CameraDevices/BackgroundReplacementDropdown.tsx
```diff
@@ -0,0 +1,79 @@
...
+      } else if (selectReplacement === ReplacementOptions.Beach && isBackgroundReplacementSupported) {
+        const imageInBase64 = BackgroundImageEncoding();
+        const blob = await createImageBlob(imageInBase64);
+        await changeBackgroundReplacementImage(blob);
+        console.info(`Video filter changed to Relacement - Beach`);
+      }      
+      // Update the current selected transform.
+      setBackgroundReplacementOption(selectReplacement);  
+    } catch (e) {
+      console.error('Error trying to apply', selectReplacement, e);
```
xuesichao, line 62: nit: Should we use `logger` from SDK instead?

apps/meeting/src/providers/AppStateProvider.tsx
```diff
@@ -26,6 +27,7 @@ interface AppStateValue {
   keepLastFrameWhenPaused: boolean;
   layout: Layout;
   joinInfo: JoinMeetingInfo | undefined;
+  backgroundReplacementOption: string;
```
xuesichao, line 30: better to make it enum or string literal instead of string.

apps/meeting/src/providers/AppStateProvider.tsx
```diff
@@ -79,30 +82,22 @@ export function AppStateProvider({ children }: Props) {
   const [videoTransformCpuUtilization, setCpuPercentage] = useState(VideoFiltersCpuUtilization.CPU40Percent);
   const [imageBlob, setImageBlob] = useState<Blob | undefined>(undefined);
   const [skipDeviceSelection, setSkipDeviceSelection] = useState(false);
+  const [backgroundReplacementOption, setBackgroundReplacementOption] = useState<string>(ReplacementOptions.Blue);
```
xuesichao, line 85: Should use <enum> or string literal instead.

apps/meeting/src/utils/image.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+    const blob = await response.blob();
+    return blob;
+  } catch (e) {
+    console.error(`Cannot create image blob with ${image}:`, e);
+    throw e;
+  }
+}
+
+export function BackgroundImageEncoding() {
```
xuesichao, line 51: Could we use webpack server to load jpg file as assets?
Something like this: https://v4.webpack.js.org/loaders/file-loader/
We might be using webpack v5 now.

apps/meeting/src/components/DeviceSelection/CameraDevices/BackgroundReplacementDropdown.tsx
dingyishen-amazon: Yeah, we could hide `BackgroundReplacementDropdown` if it is not supported.

apps/meeting/src/components/DeviceSelection/CameraDevices/BackgroundReplacementDropdown.tsx
dingyishen-amazon: If we do not handle the if-else logic here, we need to handle it in the createBlob, it will be the same case if we plan to add new options.

xuesichao: Right, anyway we will need to handle it. I think it's better to handle it in a lowerlevel, such as createBlob.
I'd hope we can simplify the component layer interface to something like below:
```
const options = [
{
  label: 'Beach'
  type: 'image'
  value: 'image path or url'
 },
 {
   label: 'Blue',
   type: 'color',
   value: '#0000ff'
 }
]
```

And we when creating the blob, we can just check the type and create image blob or color blob separately.

apps/meeting/src/utils/image.ts
```diff
@@ -0,0 +1,61 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export async function createColorBlob(): Promise<Blob> {
```
xuesichao, line 4: Could we add a parameter for `hexColor: string`. So this API can create color blob based on input?

xuesichao: Could refer to this? https://codepen.io/xuesichao/pen/GRbvpOZ
I think we can remove the gradient to make it simple.

apps/meeting/src/providers/AppStateProvider.tsx
dingyishen-amazon: may I know what id the benefit of that? The above params are using string as type.

apps/meeting/src/types/index.ts
```diff
@@ -67,6 +67,23 @@ export type VideoTransformDropdownOptionType = {
   value: string;
 };
 
+// Bcakground Replacement Options
+export const ReplacementOptions = {
```
xuesichao, line 71: Could we change this to enum as well?

```
export enum ReplacementOptions {
  Blue = 'Blue';
  Beach = 'Beach';
}
```

apps/meeting/src/utils/image.ts
```diff
@@ -0,0 +1,73 @@
...
+export async function createImageBlob(image: string): Promise<Blob> {
+  let option;
+  switch (image) {
+    case ReplacementOptions.Beach:
+      option = beach;
+    default:
+      break;
+  }
+  if (!option) {
+    throw new Error(`Unsupported replacement image: ${image}`);
```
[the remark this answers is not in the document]
xuesichao: nit: 3. if we can change image to a enum type then typescript will take care of such input validation

xuesichao (approved):

PR #449, opened 2026-06-22
fix(meeting-demo): persist selected video transform from device setup into the meeting

aryanbakshi25:
## What
  Persist the background video transform (Blur/Replacement) chosen on the
  device-setup page so it remains applied after joining the meeting.
  
  ## Why
  The in-meeting VideoInputTransformControl reset the video input to the plain
  camera on mount because it had no knowledge of what was selected earlier (this
  was an explicit TODO in the code). As a result, a filter chosen during device
  setup was silently lost on join.
  
  ## How
  - Add `activeVideoTransformOption` to AppStateProvider as a shared source of truth.
  - VideoTransformDropdown (device page) reads/writes it instead of local state.
  - VideoInputTransformControl (in-meeting) reads it and only resets to the
    intrinsic device when no transform is selected.
  
  ## Testing
  - `npm run build:fast` passes.
  - Manually verified Blur and Replacement now persist from setup into the meeting;
    selecting None is unchanged.
  
  ## Scope
  Demo app only (`apps/meeting`). No changes to the component library or JS SDK.

### DIFF
--- a/apps/meeting/src/components/MeetingControls/VideoInputTransformControl.tsx
+++ b/apps/meeting/src/components/MeetingControls/VideoInputTransformControl.tsx
@@ -48,7 +48,7 @@ const VideoInputTransformControl: React.FC<Props> = ({
   const [isLoading, setIsLoading] = useState(false);
   const [dropdownWithVideoTransformOptions, setDropdownWithVideoTransformOptions] = useState<ReactNode[] | null>(null);
   const videoDevices: DeviceType[] = useMemoCompare(devices, (prev: DeviceType[] | undefined, next: DeviceType[] | undefined): boolean => isEqual(prev, next));
-  const { backgroundReplacementOption, setBackgroundReplacementOption, replacementOptionsList, activeVideoTransformOption, setActiveVideoTransformOption } = useAppState();
+  const { backgroundReplacementOption, setBackgroundReplacementOption, replacementOptionsList, videoTransformOption, setVideoTransformOption } = useAppState();
 
   useEffect(() => {
     maybeResetDeviceToIntrinsic();
@@ -60,7 +60,7 @@ const VideoInputTransformControl: React.FC<Props> = ({
   const maybeResetDeviceToIntrinsic = async () => {
     try {
       if (
-        activeVideoTransformOption === VideoTransformOptions.None &&
+        videoTransformOption === VideoTransformOptions.None &&
         isVideoTransformDevice(selectedDevice)
       ) {
         const intrinsicDevice = await selectedDevice.intrinsicDevice();
@@ -91,7 +91,7 @@ const VideoInputTransformControl: React.FC<Props> = ({
         await current.stop();
         current = intrinsicDevice;
         // Switch to background blur device if old selection was background replacement otherwise switch to default intrinsic device.
-        if (activeVideoTransformOption === VideoTransformOptions.Replacement) {
+        if (videoTransformOption === VideoTransformOptions.Replacement) {
           current = await createBackgroundBlurDevice(current) as VideoTransformDevice;
           logger.info(`Video filter was turned on - video transform device: ${JSON.stringify(current)}`);
         } else {
@@ -108,8 +108,8 @@ const VideoInputTransformControl: React.FC<Props> = ({
       }
 
       // Update the current selected transform.
-      setActiveVideoTransformOption((activeVideoTransformOption) =>
-        activeVideoTransformOption === VideoTransformOptions.Blur
+      setVideoTransformOption((videoTransformOption) =>
+        videoTransformOption === VideoTransformOptions.Blur
           ? VideoTransformOptions.None
           : VideoTransformOptions.Blur
       );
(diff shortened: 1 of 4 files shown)


apps/meeting/src/providers/AppStateProvider.tsx
```diff
@@ -38,6 +39,10 @@ interface AppStateValue {
   backgroundReplacementOption: ReplacementOptions;
   replacementOptionsList: ReplacementDropdownOptionType[];
   enableMaxContentShares: boolean;
+  /** The video transform (None/Blur/Replacement) selected before joining, so the
+   * in-meeting control can re-apply it instead of resetting to the plain device. */
+  activeVideoTransformOption: string;
```
xuesichao: Could we make it a enum?
below `const` can be `enum` so here string can be `VideoTransformOptions`
```
// Video Transform Options
export const VideoTransformOptions = {
  None: 'None',
  Blur: 'Background Blur',
  Replacement: 'Background Replacement',
};
```

apps/meeting/src/components/MeetingControls/VideoInputTransformControl.tsx
```diff
@@ -47,21 +47,22 @@ const VideoInputTransformControl: React.FC<Props> = ({
   const { isBackgroundReplacementSupported, createBackgroundReplacementDevice, changeBackgroundReplacementImage, backgroundReplacementProcessor } = useBackgroundReplacement();
   const [isLoading, setIsLoading] = useState(false);
   const [dropdownWithVideoTransformOptions, setDropdownWithVideoTransformOptions] = useState<ReactNode[] | null>(null);
-  const [activeVideoTransformOption, setActiveVideoTransformOption] = useState<string>(VideoTransformOptions.None);
   const videoDevices: DeviceType[] = useMemoCompare(devices, (prev: DeviceType[] | undefined, next: DeviceType[] | undefined): boolean => isEqual(prev, next));
-  const { backgroundReplacementOption, setBackgroundReplacementOption, replacementOptionsList } = useAppState();
+  const { backgroundReplacementOption, setBackgroundReplacementOption, replacementOptionsList, activeVideoTransformOption, setActiveVideoTransformOption } = useAppState();
```
xuesichao: What's difference between `backgroundReplacementOption, setBackgroundReplacementOption` and `activeVideoTransformOption`? Is it possible to merge them?

apps/meeting/src/providers/AppStateProvider.tsx
xuesichao: Also feel like we can remove "Active" prefix

aryanbakshi25: enum revision done in 427d38d - converted VideoTransformOptions to a string enum and typed the field as VideoTransformOptions

rename revision done in 427d38d - renamed to videoTransformOption / setVideoTransformOption instead of activeVideoTransformOption

apps/meeting/src/components/MeetingControls/VideoInputTransformControl.tsx
```diff
@@ -47,21 +47,22 @@ const VideoInputTransformControl: React.FC<Props> = ({
   const { isBackgroundReplacementSupported, createBackgroundReplacementDevice, changeBackgroundReplacementImage, backgroundReplacementProcessor } = useBackgroundReplacement();
   const [isLoading, setIsLoading] = useState(false);
   const [dropdownWithVideoTransformOptions, setDropdownWithVideoTransformOptions] = useState<ReactNode[] | null>(null);
-  const [activeVideoTransformOption, setActiveVideoTransformOption] = useState<string>(VideoTransformOptions.None);
   const videoDevices: DeviceType[] = useMemoCompare(devices, (prev: DeviceType[] | undefined, next: DeviceType[] | undefined): boolean => isEqual(prev, next));
-  const { backgroundReplacementOption, setBackgroundReplacementOption, replacementOptionsList } = useAppState();
+  const { backgroundReplacementOption, setBackgroundReplacementOption, replacementOptionsList, videoTransformOption, setVideoTransformOption } = useAppState();
```
xuesichao, line 51: Do we still need `backgroundReplacementOption` and `setBackgroundReplacementOption`? If not, let's use `videoTransformOption, setVideoTransformOption` instead. 

Looks like we just need to rename `backgroundReplacementOption` in both component and AppStateProvider to `videoTransformOption` and use it in both replacement and blur methods.

aryanbakshi25: `backgroundReplacementOption` and `setBackgroundReplacementOption` are different from `videoTransformOption` and `setVideoTransformOption`

`videoTransformOption` tells which mode is selected (None / Blur / Replacement) + checkmarks on blur/replacement
`backgroundReplacementOption` tells the actual image that background is being replaced with (Blue / Beach) and drives the `imageBlob` load + checkmarks on blue/beach. `backgroundReplacementOption` is relevant when `videoTransformOption` is set to "Replacement"

xuesichao (approved):

ltrung (approved):

PR #450, opened 2026-07-21
Add opt-in pre-meeting device setup to the meeting demo

aryanbakshi25:
# Add opt-in pre-meeting device setup to the meeting demo

## What this changes

Adds an opt-in flow that lets a user preview and select their camera and microphone on the
device-setup page before the meeting is joined, using the `deviceController` prop added to
`MeetingProvider` in [amazon-chime-sdk-component-library-react#1036](https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1036).
Selected devices persist into the meeting and across leaving and rejoining.

The feature is off by default and controlled by a new **"Set Up Devices Before Joining"** checkbox on
the join form. When unchecked, the demo's existing join-then-setup flow is unchanged.

## Flow

- The checkbox toggles the `isPreMeetingDeviceSetupAllowed` app-state flag.
- When enabled, `MeetingProviderWrapper` constructs a `DefaultDeviceController` and passes it to
  `MeetingProvider` through the `deviceController` prop. The app owns the controller and destroys it
  on unmount.
- The join form navigates to the device-setup page instead of joining, so device enumeration,
  selection, camera preview, and the microphone activity meter work before the meeting session exists.
- The meeting is joined and started from the device-setup page's "Join meeting" button.

## Changes

- `package.json` — pin `amazon-chime-sdk-js` to `3.27.1`.
- `src/providers/AppStateProvider.tsx` — add the `isPreMeetingDeviceSetupAllowed` flag and its toggle.
- `src/containers/MeetingForm/index.tsx` — add the checkbox; navigate to the device-setup page when
  opted in.
- `src/containers/MeetingProviderWrapper/index.tsx` — construct the `DefaultDeviceController` when
  opted in, pass it as the `deviceController` prop, key the provider on its presence, and destroy it
  on unmount.
- `src/views/DeviceSetup/index.tsx` — support device setup before a meeting exists.
- `src/containers/MeetingJoinDetails.tsx` — join and start the meeting from the device-setup page.
- `src/containers/NoMeetingRedirect.tsx` — skip the home redirect while in pre-meeting setup.

## Testing

- **Opted in** — device preview and selection work on the setup page before join; selected devices
  carry into the meeting and persist across leave and rejoin.
- **Opted out (default)** — controller is created at join and destroyed on leave, unchanged from the
  current demo.


apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -146,6 +148,15 @@ const MeetingForm: React.FC = () => {
...
 
+      // Set up devices before joining: go to the device-setup page and defer join/start to it.
+      // Spectators have no device setup and join immediately below.
+      if (persistDeviceController && meetingMode !== MeetingMode.Spectator) {
+        setMeetingMode(MeetingMode.Attendee);
+        navigate(routes.DEVICE);
+        setIsLoading(false);
+        return;
+      }
+
```
xuesichao: I'm not exactly sure, do we need this?
I feel like `persistDeviceController` has nothing to do with `MeetingMode.Spectator`, 

When `persistDeviceController` is true, it should defer the `mm.join()` call and call `mm.setupDevices()(or equivalent)`.

apps/meeting/src/containers/MeetingJoinDetails.tsx
```diff
@@ -22,12 +25,28 @@ const MeetingJoinDetails = () => {
...
   const handleJoinMeeting = async () => {
     setIsLoading(true);
 
     try {
+      // When devices were set up before joining, the meeting is not joined yet — join it here before
+      // starting.
+      if (persistDeviceController) {
+        if (!joinInfo) {
+          throw new Error('Missing meeting join information');
+        }
```
xuesichao: 1. I'm not 100% sure here. Technically it does not need joinInfo to use device setup page when `persistDeviceController` is true. This indicate we could move Device Setup Page before MeetingJoinForm page. On the other hand, we'd want to recommend customer to call our AWS SDK API so we can bill them.
2. We should already have this `if (!joinInfo)` guard somewhere else, don't think we need it, could you confirm?
3. why do we still call `mm.join` when `persistDeviceController` is true here?

apps/meeting/src/views/DeviceSetup/index.tsx
```diff
@@ -1,20 +1,34 @@
...
+  const meetingManager = useMeetingManager();
+  const { persistDeviceController } = useAppState();
+
+  // When set up to run before joining, enumerate and select devices here so the pickers and preview
+  // populate on this page.
+  useEffect(() => {
+    if (persistDeviceController) {
+      meetingManager.setupDevices(DeviceLabels.AudioAndVideo);
+    }
+  }, [meetingManager, persistDeviceController]);
```
xuesichao: Should we move this logic to `handleJoinMeeting` or `MeetingJoinDetails`?

apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -70,6 +70,8 @@ const MeetingForm: React.FC = () => {
     toggleEchoReduction,
```
xuesichao, line 70: I don't see a React SDK commit change, so this PR would break the build after it merges.

Also, let's hold this PR merge until the React SDK PR gets merged, and come back to update the React SDK commit in this PR.

apps/meeting/src/containers/MeetingForm/index.tsx
aryanbakshi25: separated the two. doing spectator check first. if they're not just a spectator, then we determine if they're opt-in or not.

apps/meeting/src/views/DeviceSetup/index.tsx
aryanbakshi25: used `invokeDeviceProvider` instead

apps/meeting/src/containers/MeetingJoinDetails.tsx
aryanbakshi25: 1/2. `joinInfo` is used here to build the `MeetingSessionConfiguration` for the deferred `join()`. it's not for rendering the device page, so it's only needed because the join is deferred to this point. `createMeetingAndAttendee` already happens in `MeetingForm` regardless of where join is 

3. calling `join()` earlier would set audioVideo during the device page and app would think it's in-meeting. but we want to show full device setup pre-meeting so `join()` is deferred to here.

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
```diff
@@ -12,6 +16,7 @@ import {
   useVoiceFocus,
 } from 'amazon-chime-sdk-component-library-react';
 import { useAppState } from '../../providers/AppStateProvider';
+import ConsoleLoggingEventController from '../../utils/ConsoleLoggingEventController';
```
xuesichao: I think it's fine to remove this event controller. We'd want customer to use our DefaultDeviceController or let sdk handles this.

apps/meeting/src/containers/MeetingJoinDetails.tsx
xuesichao: So this is the button to join the meeting on the device setup page. The app itself cannot navigate to this page if the joininfo is not filled in on the previous page.

Image

So I don't think we need this check here. Could you confirm?

apps/meeting/src/containers/NoMeetingRedirect.tsx
```diff
@@ -24,7 +26,8 @@ const NoMeetingRedirect: React.FC<PropsWithChildren> = ({ children }) => {
   };
 
   useEffect(() => {
-    if (!meetingManager.meetingSession) {
+    // When setting up devices before joining, a missing session is expected here, so do not redirect.
+    if (!meetingManager.meetingSession && !persistDeviceController) {
```
xuesichao: nit: let's swap the order of these two so it reads more naturally

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
xuesichao: This is really awkward. On the one hand, we want the device controller to have an event controller, and ideally it should be our DefaultEventController. But on the other hand, that makes it scoped to per session.

I think in our demo flow the DeviceController doesn't really persist across sessions, because our page requires joinInfo anyway. So let's use `DefaultEventController` instead or remove it. 

How do you think?

apps/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -70,6 +70,8 @@ const MeetingForm: React.FC = () => {
...
     setIsVoiceFocusEnabled,
     toggleMeetingJoinDeviceSelection,
+    persistDeviceController,
```
ziyiz-amzn: this is a UI state, this var name describes implementation details. I would have `IsPreMeetingDeviceSetupAllowed` or so.

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
```diff
@@ -31,12 +37,42 @@ const MeetingProviderWithDeviceReplacement: React.FC<PropsWithChildren> = ({ chi
     return Promise.resolve(nextDevice);
   };
 
+  // When opted in, the app constructs the device controller so device setup works before joining, and
+  // owns its lifecycle. Web Audio is fixed at construction (Voice Focus), so derive it from the choice.
+  const deviceController = useMemo(
```
ziyiz-amzn, line 40: I think this snippet is clearer. you can compare and decide. 

```
const deviceControllerRef = useRef<DefaultDeviceController | undefined>();

useEffect(() => {
  if (!persistDeviceController) {
    deviceControllerRef.current?.destroy();
    deviceControllerRef.current = undefined;
    return;
  }

  const dc = new DefaultDeviceController(
    logger,
    { enableWebAudio: isVoiceFocusDesired },
    undefined,
    new ConsoleLoggingEventController()
  );
  deviceControllerRef.current = dc;

  return () => {
    void dc.destroy();
  };
}, [persistDeviceController, isVoiceFocusDesired, logger]);
```

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
```diff
@@ -31,12 +37,42 @@ const MeetingProviderWithDeviceReplacement: React.FC<PropsWithChildren> = ({ chi
...
   const meetingConfigValue = {
     onDeviceReplacement: onDeviceReplacement as any,
     ...(enableMaxContentShares ? { maxContentShares: 2 } : {}),
+    ...(deviceController ? { deviceController } : {}),
   };
 
-  return <MeetingProvider {...meetingConfigValue}>{children}</MeetingProvider>;
+  // MeetingProvider creates its MeetingManager once, capturing the deviceController at that moment.
+  // Key it on the controller's presence so toggling device setup remounts the provider and rebuilds
+  // the manager with (or without) the controller.
```
ziyiz-amzn: nit: I would make comments more concise across this PR.

apps/meeting/src/containers/NoMeetingRedirect.tsx
aryanbakshi25: swapped

apps/meeting/src/containers/MeetingForm/index.tsx
aryanbakshi25: yes, `IsPreMeetingDeviceSetupAllowed` makes more sense, updated.

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
aryanbakshi25: removed it and its util, it was mainly just for verification, shouldn't be in the demo app

ziyiz-amzn (commented):
Please update description up to date. 
This is stale - "This exercises the component library's persistDeviceController option"

apps/meeting/src/containers/NoMeetingRedirect.tsx
aryanbakshi25: "The app itself cannot navigate to this page if the joininfo is not filled in on the previous page."

confirmed, yes you can only reach the device-setup page through the form flow which sets `joinInfo` before navigating anyway, so this is won't run. plus a direct URL load redirects home since the flag isn't persisted so removed the check

apps/meeting/src/containers/MeetingProviderWrapper/index.tsx
aryanbakshi25: Thanks, yeah I compared them and I think keeping `useMemo` here is fine. The library's `MeetingProvider` captures `deviceController` at construction (```
useState(() => new MeetingManager(logger,  deviceController))```), and the demo also reads it during render for the prop and the remount key. `useMemo` produces the controller synchronously during render so it's available at both points. a `useRef` set inside `useEffect` runs after the provider has already mounted and captured undefined and ref writes don't re-render so the controller would never reach the provider on the opt-in path.

aryanbakshi25: will make comments really concise and keep only if necessary though

xuesichao (dismissed):

apps/meeting/src/containers/MeetingJoinDetails.tsx
```diff
@@ -11,7 +11,10 @@ import {
   Modal,
   ModalBody,
   ModalHeader,
+  DeviceLabels,
 } from 'amazon-chime-sdk-component-library-react';
+import { MeetingSessionConfiguration } from 'amazon-chime-sdk-js';
+import { MeetingManagerJoinOptions } from 'amazon-chime-sdk-component-library-react/lib/providers/MeetingProvider/types';
```
devalevenkatesh: Not available from import at line 15?

apps/meeting/src/views/DeviceSetup/index.tsx
```diff
@@ -1,20 +1,34 @@
...
+      meetingManager.listAndSelectDevices(DeviceLabels.AudioAndVideo);
+    }
+  }, [meetingManager, isPreMeetingDeviceSetupAllowed]);
+
+  return (
+    <StyledLayout>
+      <Heading tag="h1" level={3} css="align-self: flex-start">
+        Device settings
+      </Heading>
+      <DeviceSelection />
```
devalevenkatesh, line 28: So earlier though this existed it dint show any devices since there was no way to list them before join? The only change from UI perspective is that before join now we can do this devcie selection vs one had to join first previously?

devalevenkatesh (dismissed):

apps/meeting/src/views/DeviceSetup/index.tsx
aryanbakshi25: yes, correct, now device setup is available before join fully

apps/meeting/src/containers/MeetingJoinDetails.tsx
aryanbakshi25: it is, removed the lib path and just had the import from component library

devalevenkatesh (approved):

ziyiz-amzn (approved):

